PR #5, opened 2016-02-16
ESLint 2.0, Round 1

lemonmade:
This PR does most of the required updates for ESLint 2.0, including:

- Added configurations for new rules and removed deprecated/ deleted rules (and updated the README accordingly).
- Updated the configs for the new `parserOptions` key.
- Removed the `rulesConfig` for the plugin.
- Did all the necessary dependency updates.

I omitted one important thing to keep this PR easy to review: ESLint allows us to bundle our config with our plugin, which will allow us to basically get the required packages down to 1. The config for consuming projects looks a bit worse (`{"extends": "plugin:shopify/es6"}` instead of `{"extends": "shopify"}`), but I think that's an acceptable tradeoff. If no one has any objections, I will do that work in an additional PR.

cc/ @Shopify/fed @bouk

### DIFF
--- a/packages/eslint-config-shopify/rules/stylistic-issues.js
+++ b/packages/eslint-config-shopify/rules/stylistic-issues.js
@@ -23,8 +23,10 @@ module.exports = {
   'func-names': 0,
   // Enforces use of function declarations or expressions
   'func-style': [1, 'declaration'],
+  // Blacklist certain identifiers to prevent them being used
+  'id-blacklist': 0,
   // This option enforces minimum and maximum identifier lengths (variable names, property names etc.)
-  'id-length': [1, {min: 2, properties: 'always', exceptions: ['x', 'y', 'i', 'j', '_']}],
+  'id-length': [1, {min: 2, properties: 'always', exceptions: ['x', 'y', 'i', 'j', '_', '$']}],
   // Require identifiers to match the provided regular expression
   'id-match': 0,
   // This option sets a specific tab width for your code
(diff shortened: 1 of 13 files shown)


lemonmade:
Some of the more meaningful rule updates:

* Require spreading chained method calls across multiple lines when >= 3 chained method calls.
* Require return from some Array-related callbacks, like Array#sort/map/every, etc.
* No class constructor for cases where it is either empty or simply delegates all arguments to super.
* Prefer rest parameters over dealing with `arguments`.

packages/eslint-config-shopify/rules/stylistic-issues.js
```diff
@@ -23,6 +23,8 @@ module.exports = {
   'func-names': 0,
   // Enforces use of function declarations or expressions
   'func-style': [1, 'declaration'],
+  // Blacklist certain identifiers to prevent them being used
+  'id-blacklist': [1, 'e'],
```
bouk: Why is e blacklisted?

bouk:
lgtm

packages/eslint-config-shopify/rules/stylistic-issues.js
```diff
@@ -23,6 +23,8 @@ module.exports = {
...
   // This option enforces minimum and maximum identifier lengths (variable names, property names etc.)
   'id-length': [1, {min: 2, properties: 'always', exceptions: ['x', 'y', 'i', 'j', '_']}],
```
bouk: Could you add `$` to the exceptions so `import _ from 'jquery'` isn't problematic?

packages/eslint-config-shopify/rules/stylistic-issues.js
lemonmade: Not sure, haha. No point in doing it anyway, since it's handled by the `id-length` rule.

packages/eslint-config-shopify/rules/stylistic-issues.js
lemonmade: :+1:

PR #6, opened 2016-02-17
Consolidate plugin and config

lemonmade:
This PR uses the new features in ESLint 2.0 to update the `eslint-plugin-shopify` to serve our shared config in addition to our custom rules. It also moves the react plugin and babel parser to be dependencies of the plugin. What this means is that installation is now considerably simpler — you just need to `npm install eslint eslint-plugin-shopify --save-dev` and add an `.eslintrc` file that points to the plugin's configs.

As mentioned in the previous PR, the config name (in the `eslintrc "extends"` field)  is a bit more complex — `shopify` becomes `plugin:shopify/esnext`, `shopify/react` becomes `plugin:shopify/react`, and `shopify/es5` becomes `plugin:shopify/es5`. A small price to pay for simpler setup and package structure, I think. The generator has also been updated to generate these new config options.

After this is merged I will bump to `8.0.0` and delete the `eslint-config-shopify` package, since it won't be updated anymore.

cc/ @Shopify/fed @bouk

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,19 +4,13 @@ This repo actually contains multiple projects in addition to the formal stylegui
 script/setup
 ```
 
-Make *every* required change across all repos. For a given rule change, this will often involve at least a change to `eslint-config-shopify` and to the README for this repo. You can lint and test your changes across all repos by running:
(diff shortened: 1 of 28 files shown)


bouk:
:shipit:

PR #9, opened 2016-03-01
Dependency and config updates

lemonmade:
This PR does a few minor things that will add up to a major version bump (9.0.0):

- Updates dependencies: `babel-eslint`, `eslint`, and `eslint-plugin-react` all had version bumps, and the React plugin in particular had a few breaking changes/ a new rule, which are included in this PR.

- Removes the old `config` package since it was added to the plugin.

- Moves the configuration files (`.babelrc`/ `.eslintrc`) into `package.json` when at the root, and into `.js` versions of the file otherwise. Also adds a new section/ rule to the styleguide, which indicates that we prefer this strategy for configuration files (cleans up the project root, consolidates all configuration into a single place, etc).

- Does a few other small cleanup things.

cc/ @Shopify/fed @bouk


bouk:
Lgtm

PR #13, opened 2016-03-11
Don't require flow

bouk:
@lemonmade I think this is better enforced on a per-project basis, and I find myself disabling it anyways


lemonmade:
:+1: I have manually turned this off in the majority of projects I've worked on. Minor bump?

PR #15, opened 2016-03-15
Dependency updates and Babel Plugin

lemonmade:
This PR updates a few dependencies:

- eslint (2.4.0): adds a couple of rules that aren't meaningful for us. Also enables us to write stringified versions of the severity, rather than numbers, which I much prefer and have transitioned to.

- babel-eslint (6.0.0): fixes some issues with newer versions of ESLint. It's currently in beta, I won't ship this until the stable version is out but I figured I'd share this to make it ready to go when it ships.

- eslint-plugin-react (4.2.2): adds a preference for stateless components, which I've disabled until the React test utils catch up with stateless components.

I also added eslint-plugin-babel. This plugin basically replaces a few built-in rules with customizations supporting ES2015+ features. In particular, this should resolve the weird import issue @bouk found on the first Sello conversion PR, and a problem with async arrow functions I have just been hacking around on a few different projects.

cc/ @bouk


bouk:
hmm I would've preferred it if you added eslint-plugin-babel seperately, because that's super useful (I had no idea it existed)

it all looks fine to me though

lemonmade:
@bouk can you take another look? babel-eslint hit a stable version, and I added ESLint 2.5.0 rules in 625ae7a.

bouk:
lgtm!

PR #17, opened 2016-03-18
Put all dev dependencies at the root

lemonmade:
This PR moves all the dev dependencies to the root. This makes it so we don't have to download multiple versions of some enormous packages (Babel is particularly bad on this). The tests scripts were all updated so they can be run from the individual packages, as well.

cc/ @bouk

### DIFF
--- a/script/test
+++ b/script/test
@@ -1,3 +1,16 @@
-#!/bin/bash
-set -ex
-npm run check
+#!/usr/bin/env node
+
+var path = require('path');
+var shell = require('shelljs');
+var execSync = require('child_process').execSync;
+var packageRoot = path.join(__dirname, '../packages');
+
+test('shopify-codemod');
+test('generator-eslint-shopify');
+test('eslint-plugin-shopify');
+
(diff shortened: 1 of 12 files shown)


lemonmade:
cc/ @GoodForOneFare

GoodForOneFare:
Was [linklocal](https://github.com/timoxley/linklocal) ever a consideration?  It's probably a bit too quirky, but I kinda like having an explicit list of dev dependencies per project.

script/test
```diff
@@ -1,3 +1,16 @@
-#!/bin/bash
-set -ex
-npm run check
+#!/usr/bin/env node
```
GoodForOneFare: Should this attempt to install dependencies for sub-projects?  I have to run `npm install` inside `generate-eslint-shopify` to get this working (otherwise, it spits out `Uncaught Error: Cannot find module 'yosay'`).

lemonmade: Running `script/setup` should install all dependencies, I'm assuming it's already been run before this would be run.

PR #18, opened 2016-03-18
More mods

bouk:
@lemonmade


packages/shopify-codemod/transforms/constant-function-expression-to-statement.js
```diff
@@ -0,0 +1,30 @@
...
+  const printOptions = options.printOptions || {quote: 'single'};
+
+  return j(file.source)
+    .find(j.VariableDeclaration, (p) => j.match(p, {
+      kind: "const",
+      declarations: [{
+        type: "VariableDeclarator",
+        init: {type: (type) => type == "FunctionExpression" || type == "ArrowFunctionExpression"}
+      }]
+    }) && (p.declarations[0].init.type !== "ArrowFunctionExpression" || j(p).find(j.ThisExpression).size() == 0))
```
lemonmade: `===`

lemonmade: Also, `path` > `p`

packages/shopify-codemod/transforms/constant-function-expression-to-statement.js
```diff
@@ -0,0 +1,30 @@
+export default function constantFunctionValueToStatement(file, api, options) {
+  const j = api.jscodeshift;
+  const {expression, statement, statements} = j.template;
+  const printOptions = options.printOptions || {quote: 'single'};
+
+  return j(file.source)
+    .find(j.VariableDeclaration, (p) => j.match(p, {
```
lemonmade: Didn't know you could pass a function here, is it different than using `filter`?

lemonmade:
Looks good :+1:

packages/shopify-codemod/transforms/ternary-statement-to-if-statement.js
```diff
@@ -0,0 +1,33 @@
...
+        statement = j.expressionStatement;
+        break;
+      case 'ReturnStatement':
+        p = parent.get('argument');
+        statement = j.returnStatement;
+        break;
+      }
+
+      const consequent = j.blockStatement([statement(p.node.consequent)]);
+      let alternate = null;
```
lemonmade: `let alternate;`

packages/shopify-codemod/transforms/ternary-statement-to-if-statement.js
```diff
@@ -0,0 +1,33 @@
...
+      if (!j.match(p.node.alternate, {type: 'Identifier', name: 'undefined'})) {
+        alternate = j.blockStatement([statement(p.node.alternate)]);
+      }
+
+      return j.ifStatement(p.node.test, consequent, alternate);
+    })
+    .toSource(printOptions);
+};
```
lemonmade: README updates for these would be good too.

packages/shopify-codemod/transforms/constant-function-expression-to-statement.js
bouk: well, with filter I'd also have to do `.type === 'VariableDeclaration'`. This is what gets called from find https://github.com/facebook/jscodeshift/blob/master/src/matchNode.js

packages/shopify-codemod/transforms/ternary-statement-to-if-statement.js
bouk: well, `let alternate;` would make `alternate` `undefined` if I don't assign it, which is kind of bad, because it should be null. Small difference

lemonmade: Yeah, I guess it's not a big deal either way, I just like the consistency with the other `let`s but I get that it's different since it might not be assigned at all.

PR #19, opened 2016-03-18
Separate function to arrow transform

bouk:
@lemonmade

### DIFF
--- a/packages/shopify-codemod/test/transforms/function-to-arrow.test.js
+++ b/packages/shopify-codemod/test/transforms/function-to-arrow.test.js
@@ -1,0 +1,12 @@
+import 'test-helper';
+import ternaryStatementToIfStatement from 'function-to-arrow';
+
+describe('functionToArrow', () => {
+  it('transforms the simple case', () => {
+    expect(ternaryStatementToIfStatement).to.transform('function-to-arrow/basic');
+  });
+
+  it("doesn't transform function that use `this`", () => {
+    expect(ternaryStatementToIfStatement).to.transform('function-to-arrow/this');
+  });
+});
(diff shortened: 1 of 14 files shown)


packages/shopify-codemod/test/transforms/function-to-arrow.test.js
```diff
@@ -0,0 +1,12 @@
...
+import ternaryStatementToIfStatement from 'function-to-arrow';
+
+describe('functionToArrow', () => {
+  it('transforms the simple case', () => {
+    expect(ternaryStatementToIfStatement).to.transform('function-to-arrow/basic');
+  });
+
+  it("doesn't transform function that use `this`", () => {
+    expect(ternaryStatementToIfStatement).to.transform('function-to-arrow/this');
+  });
```
lemonmade: Can you add a function that it still converts a function that has a nested `this` in a callback? Something like:

```js
const foo = function() {
  bar(function() {
    this.baz;
  });
}
```

should become:

```js
const foo = () => {
  bar(function() {
    this.baz;
  });
}
```

lemonmade: Although this will probably make the mod a fair bit more complex.

bouk: that specific example actually gets turned into

```
function foo() {
  bar(function() {
    this.baz;
  });
}
``

:smile:

bouk: but I don't think we should support that specific case, probably doesn't happen too often and it's more trouble than it's worth

lemonmade: :+1:

lemonmade: Add docs!

PR #20, opened 2016-03-18
Add codemod to remove unneeded returns from test files

bouk:
@lemonmade this will save a lot of time when translating test files

### DIFF
--- a/packages/shopify-codemod/transforms/remove-useless-return-from-test.js
+++ b/packages/shopify-codemod/transforms/remove-useless-return-from-test.js
@@ -1,0 +1,50 @@
+// from https://github.com/sindresorhus/globals/blob/1e9ebc39828b92bd5c8ec7dc7bb07d62f2fb0153/globals.json#L852
+const FUNCTIONS = [
+  'after',
+  'afterEach',
+  'before',
+  'beforeEach',
+  'context',
+  'describe',
+  'it',
+  'mocha',
+  'setup',
+  'specify',
+  'suite',
+  'suiteSetup',
+  'suiteTeardown',
+  'teardown',
+  'test',
+  'xcontext',
+  'xdescribe',
+  'xit',
+  'xspecify',
+];
+
+export default function removeUselessReturnFromTest({source}, {jscodeshift: j}, {printOptions = {}}) {
+  function matchLast(matcher) {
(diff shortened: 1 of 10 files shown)


packages/shopify-codemod/transforms/remove-useless-return-from-test.js
```diff
@@ -0,0 +1,53 @@
...
+
+  function matchLast(matcher) {
+    return (nodes) => nodes.length > 0 && j.match(nodes[nodes.length - 1], matcher);
+  }
+
+  return j(file.source)
+    .find(j.CallExpression, {
+      callee: {
+        type: 'Identifier',
+        name: (name) => FUNCTIONS.indexOf(name) >= 0,
```
lemonmade: Welp, `includes` can't come soon enough.

packages/shopify-codemod/transforms/remove-useless-return-from-test.js
```diff
@@ -0,0 +1,53 @@
...
+  const printOptions = options.printOptions || {quote: 'single'};
+
+  function matchLast(matcher) {
+    return (nodes) => nodes.length > 0 && j.match(nodes[nodes.length - 1], matcher);
+  }
+
+  return j(file.source)
+    .find(j.CallExpression, {
+      callee: {
+        type: 'Identifier',
```
lemonmade: Do you actually need to specify the type?

packages/shopify-codemod/transforms/remove-useless-return-from-test.js
```diff
@@ -0,0 +1,53 @@
...
+  'test',
+  'xcontext',
+  'xdescribe',
+  'xit',
+  'xspecify',
+];
+
+export default function removeUselessReturnFromTest(file, api, options) {
+  const j = api.jscodeshift;
+  const printOptions = options.printOptions || {quote: 'single'};
```
lemonmade: Can you copy the parameter destructuring stuff I did for the other transforms?

lemonmade:
Looks really good, reads extremely cleanly. :+1:

packages/shopify-codemod/transforms/remove-useless-return-from-test.js
bouk: yes, the callee can be something other than an identifier

PR #22, opened 2016-03-21
Add global-assignment-to-default-export and global-reference-to-import

bouk:
@lemonmade 
This automatically finds references to global references and translates them to import statements.

It also finds any global variable assignments and translates them to export statements

### DIFF
--- a/packages/shopify-codemod/README.md
+++ b/packages/shopify-codemod/README.md
@@ -55,6 +55,48 @@ a(() => {
 });
 ```
 
+### `global-assignment-to-default-export`
+
+Use magic to automatically transform global variable references into import statements.
+
+This transform is meant to be used in conjuction with [sprockets-commoner](https://github.com/Shopify/sprockets-commoner) (which is not open source at time of writing). It depends on `appGlobalIdentifiers` being set to an array of strings containing the global namespaces of your application (`App` in this example).
+
+```sh
+jscodeshift -t shopify-codemods/transforms/global-assignment-to-default-export <file>
+```
+
+#### Example
+
+```js
+App.whatever = 1;
+
+// BECOMES:
+
+'expose App.whatever';
+export default 1;
+```
+
+### `global-reference-to-import`
+
+Use magic to automatically transform global variable references into import statements.
+
+This transform is meant to be used in conjuction with [sprockets-commoner](https://github.com/Shopify/sprockets-commoner) (which is not open source at time of writing). It depends on `appGlobalIdentifiers` being set to an array of strings containing the global namespaces of your application (`App` in this example). It also requires `javascriptSourceLocation` to be set to a folder containing all the source files.
+
+```sh
+jscodeshift -t shopify-codemods/transforms/global-reference-to-import <file>
+```
+
+#### Example
+
+```js
+console.log(App.Components.UIList);
+
+// BECOMES:
+
+import UIList from 'app/components/ui_list';
+console.log(UIList);
+```
+
 ### `mocha-context-to-closure`
 
 Transforms Mocha tests that use `this` to store context shared between tests to use closure variables instead.
--- a/packages/shopify-codemod/transforms/global-assignment-to-default-export.js
+++ b/packages/shopify-codemod/transforms/global-assignment-to-default-export.js
@@ -1,0 +1,44 @@
+import {findFirstMember} from './utils';
+
+export default function globalAssignmentToDefaultExport({source}, {jscodeshift: j}, {printOptions = {}, appGlobalIdentifiers}) {
+  return j(source)
+    .find(j.Program)
+    .forEach((programPath) => {
+      let expose;
+
+      j(programPath)
+        .find(j.ExpressionStatement, {
+          expression: {
+            type: 'AssignmentExpression',
+            operator: '=',
+            left: {
+              type: 'MemberExpression',
+              object: (object) => appGlobalIdentifiers.indexOf(findFirstMember(object).name) >= 0,
+            },
+          },
+        })
+        .replaceWith((path) => {
+          const {
+            node: {
+              expression: {
+                left: member,
+                right: statement,
+              },
+            },
+          } = path;
+
+          if (expose == null) {
+            expose = j(member).toSource();
+          } else {
+            throw new Error('Found multiple exports in a single file, please break up the file first');
+          }
+          return j.exportDefaultDeclaration(statement);
+        });
+
+      if (expose != null) {
+        const {node: {body}} = programPath;
+        body.unshift(j.expressionStatement(j.literal(`expose ${expose}`)));
+      }
+    })
+    .toSource(printOptions);
+}
--- a/packages/shopify-codemod/transforms/global-reference-to-import.js
+++ b/packages/shopify-codemod/transforms/global-reference-to-import.js
@@ -1,0 +1,130 @@
+import {spawnSync} from 'child_process';
+import {resolve, relative} from 'path';
+import {findFirstMember, findLastMember} from './utils';
+
+function determineFileSearcher() {
+  if (spawnSync('which', ['ag']).status === 0) {
+    return ['ag', ['--vimgrep']];
+  } else if (spawnSync('which', ['ack']).status === 0) {
+    return ['ack', ['--column', '--no-color', '--no-group', '--with-filename']];
+  } else if (spawnSync('which', ['ack-grep']).status === 0) {
+    return ['ack-grep', ['--column', '--no-color', '--no-group', '--with-filename']];
+  } else {
+    throw new Error('global-reference-to-import: Unable to find ack or the_silver_searcher binary');
+  }
+}
+
+export default function globalReferenceToImport(
+  {source},
+  {jscodeshift: j},
+  {
+    printOptions = {},
+    javascriptSourceLocation,
+    appGlobalIdentifiers,
+  }) {
+  const [binary, args] = determineFileSearcher();
+
+  /*
+  * findDeclaringFile uses ack or the_silver_searcher to find the file in which something is declared.
+  * It can use either, because ack is ubiquitous, while the_silver_searcher is much faster (like, 10x easily).
+  */
+  function findDeclaringFile(identifier) {
+    const absolutePath = resolve(javascriptSourceLocation);
+    const escapedIdentifier = identifier.replace(/[\\^$*+?.()|[\]{}]/g, '\\$&');
+    const regex = `^(class ${escapedIdentifier}(\\s+|^)|${escapedIdentifier}\\s*=|["']expose ${escapedIdentifier}["'])`;
+    const result = spawnSync(binary, [...args, regex, absolutePath]);
+
+    if (result.error != null) {
+      throw result.error;
+    } else if (result.status === 1) {
+      return null;
+    } else if (result.status !== 0) {
+      throw new Error(`Unknown error while resolving identifier:\n${result.stderr}`);
+    }
+
+    const stdout = result.stdout.toString();
+    const parts = stdout.trim().split('\n');
+    if (parts.length === 0) {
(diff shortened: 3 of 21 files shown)


lemonmade:
cc/ @GoodForOneFare

packages/shopify-codemod/README.md
```diff
@@ -55,6 +55,44 @@ a(() => {
 });
 ```
 
+### `global-assignment-to-default-export`
```
lemonmade: I thought that commoner was doing this for us?

lemonmade: Derp, ignore.

lemonmade: You might want to mention that this is intended for use with commoner given that it uses the `expose` directive.

packages/shopify-codemod/README.md
```diff
@@ -55,6 +55,44 @@ a(() => {
...
+```
+
+#### Example
+
+```js
+console.log(App.Components.UIList);
+
+// BECOMES:
+
+import UIList from 'app/components/ui_list';
```
lemonmade: Again, might want to mention that this is really meant for our codebase more than anyone else's, in case anyone stumbles across this and wants to use it.

packages/shopify-codemod/transforms/global-assignment-to-default-export.js
```diff
@@ -0,0 +1,47 @@
+export default function globalAssignmentToDefaultExport({source}, {jscodeshift: j}, {printOptions = {}, appGlobalIdentifiers}) {
+  function findFirstMember(memberExpression) {
+    if (j.MemberExpression.check(memberExpression)) {
+      return findFirstMember(memberExpression.object);
```
lemonmade: I know in Sello we did a lot of `Sello.Components.Component = Component`. Do we care about supporting the multiple nested namespaces?

packages/shopify-codemod/transforms/global-assignment-to-default-export.js
```diff
@@ -0,0 +1,47 @@
...
+          if (expose == null) {
+            expose = j(member).toSource();
+          } else {
+            throw new Error('Found multiple exports in a single file, please break up the file first');
+          }
+          return j.exportDefaultDeclaration(statement);
+        });
+      if (expose !== null) {
+        const {node: {body}} = programPath;
+        body.unshift(j.expressionStatement(j.literal(`expose ${expose}`)));
```
lemonmade: I think there's an `insertBefore` method that would be better here.

packages/shopify-codemod/transforms/global-assignment-to-default-export.js
```diff
@@ -0,0 +1,47 @@
...
+    if (j.MemberExpression.check(memberExpression)) {
+      return findFirstMember(memberExpression.object);
+    }
+    return memberExpression;
+  }
+
+  return j(source)
+    .find(j.Program)
+    .forEach((programPath) => {
+      let expose = null;
```
lemonmade: `let expose;`

packages/shopify-codemod/transforms/global-assignment-to-default-export.js
```diff
@@ -0,0 +1,47 @@
...
+          } = path;
+
+          if (expose == null) {
+            expose = j(member).toSource();
+          } else {
+            throw new Error('Found multiple exports in a single file, please break up the file first');
+          }
+          return j.exportDefaultDeclaration(statement);
+        });
+      if (expose !== null) {
```
lemonmade: `expose != null`

lemonmade: Or should we just prefer `expose`/ `!expose`?

packages/shopify-codemod/transforms/global-reference-to-import.js
```diff
@@ -0,0 +1,144 @@
...
+  /*
+  * findDeclaringFile uses ack or the_silver_searcher to find the file in which something is declared.
+  * It can use either, because ack is ubiquitous, while the_silver_searcher is much faster (like, 10x easily).
+  */
+  function findDeclaringFile(identifier) {
+    const absolutePath = resolve(javascriptSourceLocation);
+    const escapedIdentifier = identifier.replace(/[\\^$*+?.()|[\]{}]/g, '\\$&');
+    const regex = `^(class ${escapedIdentifier}(\\s+|^)|${escapedIdentifier}\\s*=|["']expose ${escapedIdentifier}["'])`;
+    const result = useSilverSearcher
+    ? spawnSync('ag', [
```
lemonmade: Needs one level of indentation for all of this stuff.

packages/shopify-codemod/transforms/global-reference-to-import.js
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  function findFirstMember(memberExpression) {
+    if (j.MemberExpression.check(memberExpression)) {
+      return findFirstMember(memberExpression.object);
+    }
+    return memberExpression;
+  }
+
+  function findLastMember(memberExpression) {
```
lemonmade: Maybe put these in some sort of utilities module.

packages/shopify-codemod/transforms/global-reference-to-import.js
```diff
@@ -0,0 +1,144 @@
...
+      if (name == null) {
+        return;
+      }
+      j(declarationPath.get('declarations').get(0)).renameTo(name);
+      declarationPath.prune();
+    });
+
+    // Last rename any remaining global references
+    j(path)
+    .find(j.MemberExpression, {
```
lemonmade: Indentation is all off in here.

packages/shopify-codemod/transforms/global-reference-to-import.js
```diff
@@ -0,0 +1,144 @@
...
+        },
+        init: {
+          type: 'MemberExpression',
+          object: isGlobalReference,
+        },
+      }],
+      kind: 'const',
+    })
+    .forEach((declarationPath) => {
+      const {node: {declarations: [declaration]}} = declarationPath;
```
lemonmade: You can also do `declarationPath.get('declarations', 0)`

lemonmade: Which you do below so you obviously already know that :P

packages/shopify-codemod/transforms/global-reference-to-import.js
```diff
@@ -0,0 +1,144 @@
...
+        type: 'VariableDeclarator',
+        id: {
+          type: 'Identifier',
+        },
+        init: {
+          type: 'MemberExpression',
+          object: isGlobalReference,
+        },
+      }],
+      kind: 'const',
```
lemonmade: Is there a reason to restrict this to `const`?

packages/shopify-codemod/README.md
lemonmade: I see that this is not actually specific to our setup, except that it assumes you add the root path to your node path, so maybe just a note on that.

packages/shopify-codemod/transforms/global-reference-to-import.js
```diff
@@ -0,0 +1,144 @@
...
+    const stdout = result.stdout.toString();
+    const parts = stdout.trim().split('\n');
+    if (parts.length === 0) {
+      return null;
+    } else if (parts.length !== 1) {
+      throw new Error(`Found multiple definitions for ${identifier}`);
+    }
+
+    const [filename] = parts[0].split(':', 4);
+    return filename.replace(new RegExp(`^${absolutePath}/`), '').replace(/\.[a-z]+$/, '');
```
lemonmade: `path.relative` is probably a better way to get the relative path.

lemonmade: This is fabulously smart, by the way, really awesome.

bouk: TIL path.relative, thx

packages/shopify-codemod/transforms/global-reference-to-import.js
```diff
@@ -0,0 +1,144 @@
...
+      if (name == null) {
+        return memberPath.node;
+      } else {
+        return j.identifier(name);
+      }
+    });
+
+    for (const {file, name} of imports.values()) {
+      const {node: {body}} = path;
+      body.unshift(
```
lemonmade: Same comment with `insertBefore`, I forget if it's on the node or the path.

packages/shopify-codemod/transforms/global-reference-to-import.js
```diff
@@ -0,0 +1,144 @@
...
+      j(declarationPath.get('declarations').get(0)).renameTo(name);
+      declarationPath.prune();
+    });
+
+    // Last rename any remaining global references
+    j(path)
+    .find(j.MemberExpression, {
+      object: isGlobalReference,
+    })
+    .replaceWith((memberPath) => {
```
lemonmade: `{node}` since you just use it.

packages/shopify-codemod/transforms/global-assignment-to-default-export.js
bouk: We do, the `findFirstMember` is just for validating that it's a global variable assignment that we care about.

`Sello.Components.Component = Component`

would get converted to `export default Component`. I guess we could also go through the scope and find where `Component` is defined in that case, but I don't really care that much

lemonmade: Yeah, I guess we'd just put `Sello` as the `appGlobalIdentifiers`, I was thinking we'd do something like `Sello.Components` but that's unnecessary. KCSO :+1:

packages/shopify-codemod/transforms/global-assignment-to-default-export.js
bouk: it only works if you want to insert something before a node (as the name implies). The File is the top-level node, so there's no place to insert before it. Prepending the body works best in this case

lemonmade: I guess I meant to `insertBefore` the body's first expression, but I guess they have the same effect.

packages/shopify-codemod/transforms/global-reference-to-import.js
bouk: If it's not `const` then we can't translate it to a static import binding, which means it needs manual fixing.

This code mod is meant to be run after a `var2let/const` transform

lemonmade: Ah, makes sense, wasn't clear on the ordering :+1:

packages/shopify-codemod/transforms/global-reference-to-import.js
bouk: ah yeah I'll change this to also use that

packages/shopify-codemod/transforms/global-assignment-to-default-export.js
bouk: I prefer using `let expose;` when I am assigning it for sure, and `let expose = null` when I might not be assigning anything. In this case I might not assign anything to `expose`, so I assign `null`, instead of using `undefined`

bouk:
@lemonmade addressed all the comments, :ship:?

packages/shopify-codemod/transforms/global-assignment-to-default-export.js
lemonmade: I'm not a huge fan of that rule since it adds complexity to the recommendation for questionable benefit. Can we just stick to the one? Either way, should probably just be `!= null` to match above.

bouk: sure, I don't feel that strongly

PR #24, opened 2016-03-22
Convert default value assignments to use `if` statements

GoodForOneFare:
@lemonmade Please take a look!  I was tempted to convert these to `foo = (foo || 'bar')`, but the verbose Boolean check seems more in line with our conventions.  

I wasn't able to get rid of the trailing semicolons; that may be more of a recast config issue, so I'll keep digging.


packages/shopify-codemod/transforms/conditional-assign-to-if-statement.js
```diff
@@ -0,0 +1,62 @@
...
+}
+
+function negateExistenceCondition(j, oldCondition) {
+  const booleanCheck = j.callExpression(j.identifier('Boolean'), [oldCondition]);
+  const newCondition = j.unaryExpression('!', booleanCheck);
+
+  return newCondition;
+}
+
+export default function conditionalAssignToIfStatement(file, api, options) {
```
lemonmade: Can you copy the same destructing we have in the params for the other transforms?

packages/shopify-codemod/transforms/conditional-assign-to-if-statement.js
```diff
@@ -0,0 +1,62 @@
...
+}
+
+function blockifyAssignment(j, node) {
+  const assignment = j.expressionStatement(node);
+  const assignmentBlock = j.blockStatement([assignment]);
+
+  return assignmentBlock;
+}
+
+function negateExistenceCondition(j, oldCondition) {
```
lemonmade: I would just move these into the body of the main function so you don't have to worry about passing `j` around.

packages/shopify-codemod/transforms/conditional-assign-to-if-statement.js
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    aNode.type === bNode.type &&
+    aNode.name === bNode.name
+  );
+}
+
+function isAssigningToConditionObject(node) {
+  const condition = node.left;
+  const assignment = node.right.left;
+  const isMemberExpression = (condition.type === 'MemberExpression');
```
lemonmade: Prefer using `j.MemberExpression.check(condition)` for type checks.

packages/shopify-codemod/transforms/conditional-assign-to-if-statement.js
```diff
@@ -0,0 +1,62 @@
...
+  const isUsingOr = (node.operator === '||');
+  const isLeftCheckingVariable = (node.left.type === 'MemberExpression') || (node.left.type === 'Identifier');
+  const isRightAssignment = (node.right.type === 'AssignmentExpression');
+
+  return isUsingOr && isLeftCheckingVariable && isRightAssignment;
+}
+
+function isMemberEqual(aNode, bNode) {
+  return (
+    aNode.object.type === bNode.object.type &&
```
lemonmade: I feel like there should be an easier way of finding out if two nodes are equivalent but I couldn't find it :S

packages/shopify-codemod/transforms/conditional-assign-to-if-statement.js
```diff
@@ -0,0 +1,62 @@
...
+  const newCondition = j.unaryExpression('!', booleanCheck);
+
+  return newCondition;
+}
+
+export default function conditionalAssignToIfStatement(file, api, options) {
+  const j = api.jscodeshift;
+  const printOptions = options.printOptions || {quote: 'single'};
+
+  return j(file.source)
```
lemonmade: If you go up to the parent and replace on that, you can get rid of your rogue semis:

```js
  return j(file.source)
    .find(j.LogicalExpression, (node) => isConditionalAssignment(node) && isAssigningToConditionObject(node))
  	.map((path) => path.parent)
    .replaceWith((path) => {
      const logicalExpression = path.get('expression');
      const condition = negateExistenceCondition(j, logicalExpression.node.left);
      const assignment = blockifyAssignment(j, logicalExpression.node.right);

      return j.ifStatement(condition, assignment);
    })
    .toSource(printOptions);
```

packages/shopify-codemod/test/transforms/conditional-assign-to-if-statement.test.js
```diff
@@ -0,0 +1,24 @@
...
+  });
+
+  it('does not transform mismatched assignments', () => {
+    expect(conditionalAssignToIfStatement).to.transform('conditional-assign-to-if-statement/assign-to-other-member');
+  });
+
+  it('does not transform other logical operator assignments', () => {
+    expect(conditionalAssignToIfStatement).to.transform('conditional-assign-to-if-statement/assign-after-and');
+  });
+});
```
lemonmade: Another useful test to add, since this transform currently dies on it: a conditional assignment that happens as part of a return, which is pretty likely to happen given CS's implicit returns:

```coffee
foo = ->
  this.foo ||= 'lol'
```

```js
function foo() {
  return this.foo || (this.foo = 'lol');
}
```

lemonmade:
A few comments, but it's looking real good :+1:

lemonmade:
cc/ @bouk

packages/shopify-codemod/test/fixtures/conditional-assign-to-if-statement/assign-to-other-member.output.js
```diff
@@ -0,0 +1,3 @@
+this.foo || (this.bar = 'lol');
+self.foo || (self.bar = 'lols');
+x || (y = 'lols');
```
lemonmade: I'm wondering if we should just transform all of these, too — we don't really like this style, regardless of whether it's being assigned to the same thing or not.

packages/shopify-codemod/transforms/conditional-assign-to-if-statement.js
```diff
@@ -0,0 +1,62 @@
...
+  );
+}
+
+function isAssigningToConditionObject(node) {
+  const condition = node.left;
+  const assignment = node.right.left;
+  const isMemberExpression = (condition.type === 'MemberExpression');
+
+  return isMemberExpression
+    ? isMemberEqual(condition, assignment)
```
lemonmade: Will this handle multiple levels of members? `foo.bar.baz || (foo.bar.baz = 'qux');`

packages/shopify-codemod/README.md
```diff
@@ -11,12 +11,32 @@ This repository contains a collection of Codemods written with [JSCodeshift](htt
...
+```
+
+#### Example
+
+```js
+foo || (foo = 'bar');
+
+// BECOMES:
+
+if (!Boolean(foo)) {
```
bouk: You don't need to convert to boolean first if you're negating immediately

lemonmade: If we would prefer `Boolean(foo)` were there no negation, I think `!Boolean(foo)` is a better negation than `!foo`

bouk: > a better negation

what does that mean

It's very clear that in JavaScript semantics `!` convert the RHS to a boolean, so I don't see any benefit in being explicit

lemonmade: It just seems like if we prefer `!foo` here, then we should prefer `foo` or `!!foo` for the non-negated version, which we pretty explicitly don't in the styleguide.

bouk:
I think you're overcomplicating this, you could make a transform that looks for

```
.find(j.ExpressionStatement, {
  expresion: {
    type: "LogicalExpression",
    operator(op) { return ["&&", "||"].indexOf(op) >= 0; }
  }
})
```

And then transform any of those to an if statement, more generic and less complicated of a transform

packages/shopify-codemod/README.md
lemonmade: Talked it over on Slack, @GoodForOneFare let's go with the simpler `if (!foo.bar) {}` conditions. Made an issue to clarify this in the styleguide.

packages/shopify-codemod/transforms/conditional-assign-to-if-statement.js
GoodForOneFare: Done.  Thanks for the hint!

packages/shopify-codemod/transforms/conditional-assign-to-if-statement.js
GoodForOneFare: Yep.  Added this to the test to be sure.

packages/shopify-codemod/test/fixtures/conditional-assign-to-if-statement/assign-to-other-member.output.js
GoodForOneFare: This is possible, but I was thinking of doing it in another transformation.  The intent of this was to replace code generated from CS like `fakeLink ||= @document.createElement('a')`.

GoodForOneFare: That could be cleared up by having a better transform name.  `default-value-assign-to-if-statement`?

GoodForOneFare:
@bouk I haven't expanded this to transform any `&&` or `||` expression because it introduces goes beyond my original intent, and introduces a bunch of new cases (special consideration for stuff like `return parseInt(input.value, 10) || 0;` and `const ignore = (options.ignoreErrors || []).includes(xhr.status);`).

Maybe this is the basis for a broader transform, but I think I'll have to level up a bit more before I can handle that.

packages/shopify-codemod/transforms/conditional-assign-to-if-statement.js
```diff
@@ -0,0 +1,87 @@
...
+    .find(j.LogicalExpression, {
+      type: (type) => j.LogicalExpression.name === type || j.ReturnStatement.name === type,
+      operator: '||',
+      left: (left) => j.MemberExpression.check(left) || j.Identifier.check(left),
+      right: (right) => j.AssignmentExpression.check(right),
+    })
+    .filter(isAssigningToConditionObject)
+    .map((path) => path.parent)
+    .replaceWith((path) => {
+      if (j.ReturnStatement.check(path.node)) {
```
GoodForOneFare: This block of conditions is kinda icky.  I think there's a better way to structure this, but everything I've come up with just looks overengineered.

bouk:
@GoodForOneFare those wouldn't get transformed though, because you'd only be transforming expressionStatements, not return statement or variabledeclarations

bouk:
Here's a simple example: https://astexplorer.net/#/H79xYnH6PH

I think you should try to operate on the statement instead of the expression, as every kind of statement requires a different approach (think expression statement vs return statement vs declarator statement etc.). A bunch of smaller scoped transformed are easier to maintain than one transform that starts from the point of a logicalexpression and then tries to figure out what the best approach is based on the parent

GoodForOneFare:
@bouk That sounds much saner.  I'll give it a go.  Thanks!

GoodForOneFare:
Closing this; I'll resubmit with more targeted PRs.

PR #26, opened 2016-03-24
Preserve directives when adding import statement

bouk:
@lemonmade

### DIFF
--- a/packages/shopify-codemod/transforms/utils.js
+++ b/packages/shopify-codemod/transforms/utils.js
@@ -11,3 +11,14 @@ export function findLastMember(node) {
   }
   return node;
 }
+
+export function insertAfterDirectives(body, newNode) {
+  let i = 0;
+  for (;i < body.length; i++) {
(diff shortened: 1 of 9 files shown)


bouk:
This makes it so we can do global-assignment-to-default-export before global-reference-to-import and preserve the expose directive

packages/shopify-codemod/transforms/utils.js
```diff
@@ -11,3 +11,14 @@ export function findLastMember(node) {
   }
   return node;
 }
+
+export function insertAfterDirectives(body, newNode) {
+  let i = 0;
+  for (;i < body.length; i++) {
```
lemonmade: Ugh, hate this way of expressing this. Can we at least move the `let i = 0` in as the first condition?

bouk: No, because i is block scoped to for for loop otherwise.

lemonmade: As shit, yeah. Gross :( could do it with a weird `findIndex` but whatever, probably better this way.

lemonmade:
:+1:

PR #27, opened 2016-03-28
Convert this used in this-assigned functions

lemonmade:
cc/ @bouk @GoodForOneFare

### DIFF
--- a/packages/shopify-codemod/test/fixtures/mocha-context-to-closure/contextually-declared-functions.input.js
+++ b/packages/shopify-codemod/test/fixtures/mocha-context-to-closure/contextually-declared-functions.input.js
@@ -1,0 +1,12 @@
+suite('top-level', function() {
+  setup('name', () => {
+    this.foo = 'foo';
+    this.bar = function() {
+      return this.foo === 'foo';
+    }
+
+    unrelated.baz = function() {
+      return this.foo === 'mystery';
+    }
+  });
+});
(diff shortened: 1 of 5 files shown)


packages/shopify-codemod/test/fixtures/mocha-context-to-closure/contextually-declared-functions.input.js
```diff
@@ -0,0 +1,12 @@
+suite('top-level', function() {
+  setup('name', () => {
+    this.foo = 'foo';
+    this.bar = function() {
+      return this.foo === 'foo';
```
bouk: this one should be converted, only arrow functions should be

bouk: hmm maybe not. I'm not sure

lemonmade: If it's a function assigned to the mocha context, it's `this` will be the mocha context, so it's safe to covert those.

bouk: Wait yes it should be, because in this case the coffeescript code would have `=>` which is converted to arrow func, so this only needs to operate on those

GoodForOneFare:
LGTM :+1:

PR #28, opened 2016-03-28
Transform conditional assignments to `if` blocks

GoodForOneFare:
This transforms vanilla expression statements.  Conditional assignments that return a value, or that immediately call into the assigned value will be handled by subsequent PRs.

/cc @lemonmade @bouk

### DIFF
--- a/packages/shopify-codemod/README.md
+++ b/packages/shopify-codemod/README.md
@@ -1,4 +1,4 @@
-# shopify-codemods
+# shopify-codemod
 
 This repository contains a collection of Codemods written with [JSCodeshift](https://github.com/facebook/jscodeshift) that will help update our old, crusty JavaScript to nice, clean JavaScript.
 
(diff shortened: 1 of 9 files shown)


packages/shopify-codemod/README.md
```diff
@@ -213,7 +233,7 @@ suite('a', () => {
 Changes ternary statement expressions to if statements.
 
 ```sh
-jscodeshift -t shopify-codemods/transforms/ternary-statement-to-if-statement <file>
+jscodeshift -t shopify-codemod/transforms/ternary-statement-to-if-statement <file>
```
lemonmade: Thank you for fixing these!

packages/shopify-codemod/test/fixtures/conditional-assign-to-if-statement/assign-to-other-identifier.input.js
```diff
@@ -0,0 +1,7 @@
+this.foo || (this.bar = 'lol');
```
lemonmade: Not sure if out of scope, but will this work with multiple conditions? `this.foo || this.bar || (this.baz = 'qux')`

lemonmade:
One minor comment, code otherwise looks great :+1: Rebase on master will probably fix the test failure.

packages/shopify-codemod/test/fixtures/conditional-assign-to-if-statement/assign-to-other-identifier.input.js
GoodForOneFare: It spits out this:
```
if (!(this.foo || this.bar)) {
  this.baz = 'qux';
}
```

That's not pretty, but when I transform the decaffed admin/sello code, I don't see any instances of this.  My preference would be to not worry about it until we see it somewhere.

lemonmade: :+1:

lemonmade:
:ship: once you get it :green_apple:

PR #32, opened 2016-04-08
Add mochaContextToGlobalReferences codemod

lemonmade:
This PR adds a new transform to handle the renaming of some things that are injected into the test context (`spy`/ `stub`/ `clock` etc) to use a global defined for the tests instead. The codemod allows you to define these as either replacing the whole member expression (`this.clock -> testClock`) or to just replace the object in the member expression with the test global (`this.spy -> sinon.spy`).

cc/ @goodforonefare

### DIFF
--- a/packages/shopify-codemod/test/fixtures/mocha-context-to-global-references/rename.output.js
+++ b/packages/shopify-codemod/test/fixtures/mocha-context-to-global-references/rename.output.js
@@ -1,0 +1,17 @@
+suite('test', function() {
+  afterEach(function() {
+    sinon.stub(foo, 'bar');
+  });
+
+  before(function() {
+    this.testSpy = sinon.spy();
+  });
+
+  setup(function() {
+    requests = sinon.server.requests;
+  });
+
+  it('works', function() {
+    sinon.server.requests.forEach(() => assert.ok(request));
+  });
+});
--- a/packages/shopify-codemod/test/transforms/mocha-context-to-global-references.test.js
+++ b/packages/shopify-codemod/test/transforms/mocha-context-to-global-references.test.js
@@ -1,0 +1,26 @@
+import 'test-helper';
+import mochaContextToGlobalReferences from 'mocha-context-to-global-references';
+
+describe('mochaContextToGlobalReferences', () => {
+  it('replaces a contextual reference to a global reference', () => {
+    expect(mochaContextToGlobalReferences).to.transform('mocha-context-to-global-references/replace', {
+      testContextToGlobals: {
+        testClock: {
+          properties: ['clock'],
+          replace: true,
+        },
+
+        testGlobal: {
+          properties: ['shouldBeReplaced', 'mustBeReplaced'],
+          replace: true,
+        }
+      }
+    });
+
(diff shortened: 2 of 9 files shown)


packages/shopify-codemod/test/transforms/mocha-context-to-global-references.test.js
```diff
@@ -0,0 +1,28 @@
...
+      testContextToGlobals: {
+        sinon: {properties: ['spy', 'stub', 'server']},
+      },
+    });
+  });
+
+  it('replaces a contextual reference to a global reference', () => {
+    expect(mochaContextToGlobalReferences).to.transform('mocha-context-to-global-references/replace', {
+      testContextToGlobals: {
+        testClock: {
```
GoodForOneFare: If I'm reading correctly, it means, "take this new name, and use it to replace any of these existing property names".  I have to reverse the intent of the transform, and that seems counter-intuitive.

I feel it'd be clearer to have, "take this name, and replace it with this one".  e.g.,
```
{
  testContextToGlobals: [
    { from: 'clock', to: 'testClock', replace: true },
    { from: 'shouldBeReplaced', to: 'testGlobal' },
    { from: 'mustBeReplaced', to: 'testGlobal' }
  ]
}
```
... or some format where old=>new is the mapping.

packages/shopify-codemod/test/fixtures/mocha-context-to-global-references/rename.output.js
```diff
@@ -0,0 +1,17 @@
+suite('test', function() {
+  afterEach(function() {
+    sinon.stub(foo, 'bar');
+  });
+
+  before(function() {
+    this.testSpy = sinon.spy();
```
GoodForOneFare: From what I've seen of the tests, `this.spy` is going to call into a sandbox.  Is there a risk of this transform changing teardown behaviour?

GoodForOneFare:
One concern about teardowns, but this looks really good.  The config bikeshedding isn't really a blocker.

packages/shopify-codemod/test/fixtures/mocha-context-to-global-references/rename.output.js
lemonmade: That would be up the user of the codemod; for us, we'll make sure that the global we change things to is the same sandbox being injected into the test context.

packages/shopify-codemod/test/transforms/mocha-context-to-global-references.test.js
lemonmade: Good points, I see where you are coming from, but I think I'm going to stick with what I have here. 1) I think it reads better for when there are multiple properties, particularly when those properties are not being replaced, and 2) I am too lazy to update all the tests :P

packages/shopify-codemod/test/fixtures/mocha-context-to-global-references/rename.output.js
GoodForOneFare: Makes sense.

GoodForOneFare:
🚢

PR #33, opened 2016-04-11
Adds lodash plugin

### LINKED ISSUE #11: Look into adding eslint-plugin-lodash

lemonmade:
This PR fixes #11 by adding `eslint-plugin-lodash` for lodash-specific linting rules. I've used lodash a fair bit but I would definitely love feedback on this one — I feel like this might be something that different projects do differently, and my opinions might not match those of the team on some of these things. You can see all the rules here: https://github.com/wix/eslint-plugin-lodash, though I've included the "summary" description of each rule above the rule itself.

cc/ @Shopify/fed @DrewMartin @owilliams


packages/eslint-plugin-shopify/lib/config/rules/lodash.js
```diff
@@ -0,0 +1,84 @@
...
+  // Prefer identity shorthand syntax
+  'identity-shorthand': ['warn', 'never'],
+  // Prefer matches property shorthand syntax
+  'matches-prop-shorthand': ['warn', 'never'],
+  // Prefer matches shorthand syntax
+  'matches-shorthand': ['warn', 'never'],
+  // Do not use .commit() on chains that should end with .value()
+  'no-commit': 'warn',
+  // Prevent chaining syntax for single method, e.g. _(x).map().value()
+  'no-single-chain': 'warn',
```
dan-menard: How clever is this? Will it fail me if I do something like `_(x).filter(function(foo) { foo.bar == 'baz'; }).join(' ')`? Or am I just misunderstanding the `_(x).map().value()` example?

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
```diff
@@ -0,0 +1,84 @@
...
+  // Enforce a specific path style for methods like get and property: array, string, or arrays only for deep paths.
+  'path-style': ['warn', 'string'],
+  // Prefer a Lodash chain over nested Lodash calls
+  'prefer-chain': 'warn',
+  // Prefer _.compact over _.filter for only truthy values.
+  'prefer-compact': 'warn',
```
dan-menard: This one confuses me. How is the linter going to know if the members of an array are all truthy?

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
```diff
@@ -0,0 +1,84 @@
...
+  // Prefer _.map over _.forEach with a push inside.
+  'prefer-map': 'error',
+  // Prefer _.reject over filter with !(expression) or x.prop1 !== value
+  'prefer-reject': 'warn',
+  // Prefer using _.prototype.thru in the chain and not call functions in the initial value, e.g. _(x).thru(f).map(g)...
+  'prefer-thru': 'off',
+  // Prefer using array and string methods in the chain and not the initial value, e.g. _(str).split(' ')...
+  'prefer-wrapper-method': 'off',
+  // Prefer using main method names instead of aliases. (fixable)
+  'preferred-alias': 'warn',
```
tessalt: I suspect I'll be cursing this one... `.first` vs `.head` being the most irritating one.

tessalt:
:+1: looks fairly conservative.

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
DrewMartin: I think you are. When you begin a chain with `_(x)` it returns a special lodash object that allows chaining to happen. Certain operations are "final", like `join`, `first`, `last` (full list [here](https://lodash.com/docs#_)), but if you don't finish with any of them you'll just end up with a lodash object instead of your array/string/object/whatever. That's when you'd call `.value()`, so you get the actual result back. This ensures you don't needlessly chain

dan-menard: Ah, that makes sense. Thanks for the clarification 👍

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
DrewMartin: I believe it's only when you omit a second argument to `_.filter` that it becomes equivalent to `_.compact`, in which case you should just use compact

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
```diff
@@ -0,0 +1,84 @@
...
+  // Use value returned from collection methods properly.
+  'collection-method-value': 'warn',
+  // Always return a value in iteratees of Lodash collection methods that aren't forEach.
+  'collection-return': 'error',
+  // Do not use .value() on chains that have already ended (e.g. with max() or reduce()) (fixable)
+  'no-double-unwrap': 'error',
+  // Do not use superfluous arguments on Lodash methods with a specified arity.
+  'no-extra-args': 'error',
+  // Prevent chaining without evaluation via value() or non-chainable methods like max().
+  'unwrap': 'error',
```
DrewMartin: This one could potentially be annoying if we don't do a straightforward chain like `_(array).map(...).filter(...).reduce(...).values()` but like
```javascript
chain = _(array)
chain = chain.map(...).filter(...)
if chain.isEmpty()
  chain.something()
else
  chain.somethingElse()
```

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
```diff
@@ -0,0 +1,84 @@
...
+
+  // Stylistic issues
+
+  // Enforce a specific chain style: explicit, implicit, or explicit only when necessary.
+  'chain-style': ['warn', 'explicit'],
```
DrewMartin: Implicit over explicit chaining IMO

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
```diff
@@ -0,0 +1,84 @@
...
+  // Prefer identity shorthand syntax
+  'identity-shorthand': ['warn', 'never'],
```
DrewMartin: Just to double check because their docs don't make it clear: this will warn when you do something like
`_.max(scores, function(score) {return score;})`
And do nothing when you do
`_.max(scores)`
right?

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
dan-menard: Examples please! What is implicit chaining?

DrewMartin: Implicit: `_(thing).map(...).reduce(...)`
Explicit: `_.chain(thing).map(...).reduce(...)`

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
```diff
@@ -0,0 +1,84 @@
...
+  // Prefer matches property shorthand syntax
+  'matches-prop-shorthand': ['warn', 'never'],
+  // Prefer matches shorthand syntax
+  'matches-shorthand': ['warn', 'never'],
```
DrewMartin: I prefer `always` for this:
`result = _.filter(users, {age: 30, name: 'Bob'}));` :+1: 
`_.filter(users, function (user) { return user.age === 30 && user.name === 'Bob'; });` :-1:

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
lemonmade: The reason I prefer explicit here is that 1) it's not obvious that the lodash object is itself a function that's usable, and if you import that object, you bring in all of lodash (whereas, we can eventually get to a point where we only import the methods from lodash we need and leave the rest of them out of the bundle, if we bring in only the functions themselves).

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
lemonmade: Correct, also warns on `_.max(scores, _.identity)` if I'm understanding the docs correctly.

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
lemonmade: Do you think it's annoying enough to warrant turning off? I kind of like not having to deal with aliases, but not sure if it's worth it in practice.

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
lemonmade: I haven't seen places we do too much of this, do you know if we do?

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
```diff
@@ -0,0 +1,84 @@
...
+  'preferred-alias': 'warn',
+  // Use/forbid property shorthand syntax.
+  'prop-shorthand': ['warn', 'never'],
+
+  // Preference over native
+
+  // Prefer _.constant over functions returning literals.
+  'prefer-constant': 'off',
+  // Prefer using _.get or _.has over expression chains like a && a.b && a.b.c.
+  'prefer-get': 'off',
```
DrewMartin: Not warn for this? I feel like it will be useful to use `.get` when transitioning from coffeescript to ES for all those places we have `a?.b?.c` style stuff

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
DrewMartin: Ehh probably not. I like chaining in theory, but I avoided it with coffeescript because it looks pretty gross with coffeescript-style brace-free functions. I don't think there's much of it right now and if we find this too restricting we can always turn it off

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
DrewMartin: Ahh importing is a good reason then :+1:

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
lemonmade: Yeah, I think you're probably right on this.

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
lemonmade: Honestly, I didn't even know this as a feature that was available. Considering I'd rather people just use the native `filter`, I don't feel too strongly either way on this.

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
dan-menard: I don't have an opinion either way, but fwiw I don't remember ever seeing `a?.b` in admin coffeescript.

lemonmade: We use the question mark operator in many places in admin.

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
owilliams: A lot of the things that had to be updated in the lodash v3 upgrade were aliases that had been removed, so my personal preference would be to force using main methods.

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
```diff
@@ -0,0 +1,84 @@
+// see https://github.com/wix/eslint-plugin-lodash
+
+module.exports = {
+  // Possible errors
+
+  // Use or avoid thisArg for Lodash method callbacks, depending on major version.
+  'callback-binding': 'warn',
```
owilliams: Since this is removed in v4 how about we make it `error`?

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
```diff
@@ -0,0 +1,84 @@
...
+  // Prevent chaining syntax for single method, e.g. _(x).map().value()
+  'no-single-chain': 'warn',
+  // Enforce a specific path style for methods like get and property: array, string, or arrays only for deep paths.
+  'path-style': ['warn', 'string'],
+  // Prefer a Lodash chain over nested Lodash calls
+  'prefer-chain': 'warn',
+  // Prefer _.compact over _.filter for only truthy values.
+  'prefer-compact': 'warn',
+  // Prefer _.filter over _.forEach with an if statement inside.
+  'prefer-filter': 'warn',
```
owilliams: Filter is typically much clearer, thoughts on `error` here?

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
owilliams: Also, it would be preferable to use consistent methods imho.

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
lemonmade: I think our general approach on error versus warn is, if this is probably a coding error or an absolutely horrendous style thing (thing `with`/ labels/ `eval`), that's an error. If it's just a bad practice or something we want to dissuade, that's a warn. IMO, this probably falls into the second category.

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
lemonmade: Just going to switch this one.

packages/eslint-plugin-shopify/lib/config/rules/lodash.js
owilliams: 👍

lemonmade:
Reorganized how you will actually get this in your project: instead of it being on automatically, you have to extend an additional config. Added explanations and such to the main README and to this package's README. Basically, looks like this:

```js
{
  "eslintConfig": {
    "extends": [
      "plugin:shopify/esnext",
      "plugin:shopify/lodash"
    ],
}
```

PR #34, opened 2016-04-12
Don't try to transform non-Mocha functions

GoodForOneFare:
Previously, `handleContext` would try to transform *any* function inside a suite.  This broke when suites invoked non-Mocha functions (e.g., `itDoesNotRemoveTheErrorClass` in `error_remover_test.coffee`).

/cc @lemonmade


lemonmade:
👍

PR #35, opened 2016-04-12
Fix function to arrow

GoodForOneFare:
Fixes `function-to-arrow` by:
- Ignoring class methods
  - (I know that `transform-class-properties` is enabled, but, eh, stage 1)
- Handling functions with argument-less `return`s

/cc @lemonmade

### DIFF
--- a/packages/shopify-codemod/transforms/function-to-arrow.js
+++ b/packages/shopify-codemod/transforms/function-to-arrow.js
@@ -1,7 +1,8 @@
 export default function functionToArrow({source}, {jscodeshift: j}, {printOptions = {}}) {
   return j(source)
     .find(j.FunctionExpression)
-    .filter((path) => j(path).find(j.ThisExpression).size() === 0 && !j.Property.check(path.parent.node))
+    .filter(({parent}) => !j.MethodDefinition.check(parent.node) && !j.Property.check(parent.node))
+    .filter((path) => j(path).find(j.ThisExpression).size() === 0)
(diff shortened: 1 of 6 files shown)


packages/shopify-codemod/transforms/function-to-arrow.js
```diff
@@ -1,14 +1,19 @@
 export default function functionToArrow({source}, {jscodeshift: j}, {printOptions = {}}) {
   return j(source)
     .find(j.FunctionExpression)
-    .filter((path) => j(path).find(j.ThisExpression).size() === 0 && !j.Property.check(path.parent.node))
+    .filter(({parent}) => !j.MethodDefinition.check(parent.node) && !j.Property.check(parent.node))
+    .filter((path) => j(path).find(j.ThisExpression).size() === 0)
```
lemonmade: Can we combine these into a single `filter`? You can still have separate functions for each check and just `&&` them together in the filter:

```js
function isConvertableFunction(path) {
  return !isMember(path) && !containsThisExpression(path)
}

.filter(isConvertibleFunction)
```

lemonmade:
One minor comment, otherwise LGTM 👍

packages/shopify-codemod/transforms/function-to-arrow.js
GoodForOneFare: That's a really nice suggestion.  Much easier to read.  Done.

lemonmade:
Can we ship this?

PR #36, opened 2016-04-12
Strip argless returns if they're the last line of a Mocha function

GoodForOneFare:
Gets rid of `return;` lines from Mocha functions (these were breaking tests like `amount_validation_test.coffee`).

/cc @lemonmade

### DIFF
--- a/packages/shopify-codemod/test/fixtures/remove-useless-return-from-test/basic.input.js
+++ b/packages/shopify-codemod/test/fixtures/remove-useless-return-from-test/basic.input.js
@@ -1,4 +1,7 @@
 suite('a', () => {
+  beforeEach(function() {
+    return;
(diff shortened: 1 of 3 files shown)


lemonmade:
Code looks good. Do you think we should add another transform to kill the empty-function-passed-to-mocha-hook situation?

GoodForOneFare:
@lemonmade RE the empty-function-passed-to-mocha-hook situation.  Do you mean the this test data?

```
  beforeEach(function() {
    return;
  });
```

If so, I don't think that'll crop up in real Shopify code.  If that's the case, I'll update the test data to be less brainitchy.

lemonmade:
Ah, I gotcha, no worries. 🚢 away.

PR #37, opened 2016-04-12
Add some ESLint rules

lemonmade:
This PR adds two ESLint rules:

1. `jquery-dollar-sign-reference` forces you to prefix references with `$` when they definitely refer to jQuery objects (and prevents you from prefixing with `$` if it definitely does not refer to a jQuery object).

2. `prefer-early-return` forces you to use an early return instead of nesting an entire function body in a conditional expression without an alternate.

Let me know if you want me to break this into two PRs, I figured I'd add them both at once but we can at list split it for ease of review.

cc/ @GoodForOneFare @bouk

### DIFF
--- a/packages/eslint-plugin-shopify/lib/rules/jquery-dollar-sign-reference.js
+++ b/packages/eslint-plugin-shopify/lib/rules/jquery-dollar-sign-reference.js
@@ -25,16 +25,16 @@ var nonChainingMethods = {
   val: true,
   width: true,
 
-  html: function(node) {
-    return node.arguments.length === 0;
+  attr: function(node) {
+    return node.arguments.length < 2 && node.arguments[0] && node.arguments[0].type !== 'ObjectExpression';
   },
   data: function(node) {
     return node.arguments.length < 2 && node.arguments[0] && node.arguments[0].type !== 'ObjectExpression';
   },
-  prop: function(node) {
-    return node.arguments.length < 2 && node.arguments[0] && node.arguments[0].type !== 'ObjectExpression';
+  html: function(node) {
+    return node.arguments.length === 0;
   },
-  attr: function(node) {
+  prop: function(node) {
     return node.arguments.length < 2 && node.arguments[0] && node.arguments[0].type !== 'ObjectExpression';
   },
 };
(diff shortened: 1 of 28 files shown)


lemonmade:
Fixes #14.

packages/eslint-plugin-shopify/lib/rules/jquery-dollar-sign-reference.js
```diff
@@ -0,0 +1,152 @@
...
+
+    checkForValidjQueryReference(node, fakeIdentifier, node.value);
+  }
+
+  return {
+    VariableDeclarator: handleVariableDeclarator,
+    AssignmentExpression: handleAssignmentExpression,
+    ObjectExpression: handleObjectExpression,
+    ClassProperty: handleClassProperty,
+    // Program: function(node) { console.log(node); }
```
GoodForOneFare: Can this commented line be removed?

packages/eslint-plugin-shopify/lib/rules/jquery-dollar-sign-reference.js
```diff
@@ -0,0 +1,152 @@
...
+  serialize: true,
+  serializeArray: true,
+  size: true,
+  text: true,
+  toArray: true,
+  triggerHandler: true,
+  val: true,
+  width: true,
+
+  html: function(node) {
```
GoodForOneFare: I guess this block of functions should be in alphabetical key order?  To be consistent with the above block :)

packages/eslint-plugin-shopify/lib/rules/jquery-dollar-sign-reference.js
```diff
@@ -0,0 +1,152 @@
...
+    case 'Identifier': return node.name;
+    case 'Literal': return node.value;
+    case 'MemberExpression': return getFinalReferenceName(node.property);
+    default: return null;
+    }
+  }
+
+  function isjQueryReference(node) {
+    var referenceName = getFinalReferenceName(node);
+    return referenceName != null && JQUERY_IDENTIFIER_REGEX.test(referenceName);
```
GoodForOneFare: `test` will happily reject a `null` input, so this could be shorter.  No big deal either way, though.

GoodForOneFare:
Inspirational code makes for uninspiring review comments 👍

packages/eslint-plugin-shopify/lib/rules/jquery-dollar-sign-reference.js
lemonmade: Did not know this :)

PR #40, opened 2016-04-13
Use a range-generating function rather than CoffeeScript output

lemonmade:
This PR adds a transform that changes all of the gross CoffeeScript-range-generated garbage into a reference to a global function. This transform will need to happen before the global reference to import one when run via edify. I couldn't think of a nice way to make this generic (so it could be used by non-Shopify code), but I figured that wasn't a big deal for now.

cc/ @GoodForOneFare


GoodForOneFare:
Damn, you're fast 👍

PR #41, opened 2016-04-15
Additional ESLint rules

lemonmade:
I had a bunch of ideas yesterday so I just built them all. Just get to this the you can :) @GoodForOneFare

### DIFF
--- a/packages/eslint-plugin-shopify/docs/rules/class-property-semi.md
+++ b/packages/eslint-plugin-shopify/docs/rules/class-property-semi.md
@@ -26,4 +26,4 @@ The success/ failure cases are reversed when the argument is explicitly `'never'
 
 ## When Not To Use It
 
-If you do not wish to enforce semicolons for class proeprties, then you can safely disable this rule.
+If you do not wish to enforce semicolons for class properties, then you can safely disable this rule.
--- a/packages/eslint-plugin-shopify/docs/rules/class-property-semi.md
+++ b/packages/eslint-plugin-shopify/docs/rules/class-property-semi.md
@@ -1,0 +1,29 @@
+# Requires (or disallows) semicolons for class properties. (class-property-semi)
+
+## Rule Details
+
+This rule takes one argument. If it is `'always'` (default) then it warns when it does not find a semicolon at the end of a class property. If `'never'` then it warns if a semicolon is present.
+
+The following patterns are considered warnings when using the default or explicitly setting the argument to `'always'`:
+
+```js
+class Foo {
+  bar = true
+  static baz = false
+}
+```
+
+The following patterns are not warnings when using the default or explicitly setting the argument to `'always'`:
+
+```js
+class Foo {
+  bar = true;
+  static baz = false;
+}
+```
+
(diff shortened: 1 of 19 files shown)


packages/eslint-plugin-shopify/docs/rules/class-property-semi.md
```diff
@@ -0,0 +1,29 @@
...
+  bar = true;
+  static baz = false;
+}
+```
+
+The success/ failure cases are reversed when the argument is explicitly `'never'`.
+
+## When Not To Use It
+
+If you do not wish to enforce semicolons for class proeprties, then you can safely disable this rule.
```
GoodForOneFare: Typo: `proeprties `

packages/eslint-plugin-shopify/docs/rules/restrict-full-import.md
```diff
@@ -0,0 +1,38 @@
+# Prevents importing the entirety of a package. (restrict-full-import)
+
+Importing the entirety of a large module can be undesirable because it becomes harder to track what properties are being used.
+
+## Rule Details
+
+This rule aims to restrict the imports of specified modules to only be an import of specific properties. This does **not** mean that you are not importing from the module, it simply means that you only extract the properties you need.
```
GoodForOneFare: The first sentence isn't easy to read.  Why is it just aiming?  I'd prefer something like, "This rule prevents default imports for a configurable list of modules."

Second sentence's "does not mean that you are not" would be easier to parse without the double negative, e.g.,: "This is **not** a module blacklist; individual properties can still be imported."

packages/eslint-plugin-shopify/docs/rules/restrict-full-import.md
```diff
@@ -0,0 +1,38 @@
...
+
+var _ = require('something-else');
+var chain = require('lodash').chain;
+var chain = require('lodash/chain');
+var {chain} = require('lodash');
+```
+
+## When Not To Use It
+
+If you do not want to restrict from full imports from modules, you can safely disable this rule.
```
GoodForOneFare: Reword: `restrict from full`.

packages/eslint-plugin-shopify/lib/rules/class-property-semi.js
```diff
@@ -0,0 +1,41 @@
...
+        fix: function(fixer) {
+          fixer.insertTextAfter(lastToken, ';');
+        },
+      });
+    } else if (!always && hasSemicolon) {
+      context.report({
+        node: node,
+        message: 'Extra semicolon.',
+        fix: function(fixer) {
+          fixer.remove(lastToken);
```
GoodForOneFare: Whoa! Nice!

GoodForOneFare:
Quibbles over wording, but the code is solid 👍

packages/eslint-plugin-shopify/docs/rules/restrict-full-import.md
lemonmade: Good call 👍

PR #44, opened 2016-04-18
Add prefer-class-properties rule

lemonmade:
One more rule! Also updated some docs for all the new rules.

cc/ @GoodForOneFare

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ This repository contains everything you should need for writing JavaScript at Sh
 1. [Strings](#strings)
 1. [Functions](#functions)
 1. [Types and casting](#types-and-casting)
-1. [ES2015 features](#es2015-features)
+1. [ESNext features](#esnext-features)
 1. [Project structure](#project-structure)
 1. [Resources](#resources)
 
@@ -212,7 +212,7 @@ npm run lint
 
 - [3.1](#3.1) <a name="3.1"></a> Always use semicolons.
 
-  ESLint rule: [`semi`](http://eslint.org/docs/rules/semi.html)
+  ESLint rules: [`semi`](http://eslint.org/docs/rules/semi.html) and [`class-property-semi`](packages/eslint-plugin-shopify/docs/class-property-semi.md)
 
   ```js
   // bad
@@ -799,6 +799,29 @@ npm run lint
   }
 
   const good = calculateGood(someCondition, someOtherCondition, someFinalCondition);
+  ```
+
+- [6.5](#6.5) <a name="6.5"></a> Don’t nest an entire function body inside a conditional. Instead, return early using the opposite of the conditional.
+
+  > Why? This reduces indentation and makes the code easier to read.
+
+  ESLint rule: [`prefer-early-return`](packages/eslint-plugin-shopify/docs/prefer-early-return.md)
+
+  ```js
+  // bad
+  function badFunc() {
+    if (something) {
+      doThis();
+      andThat();
+    }
+  }
+
+  // good
+  function goodFunc() {
+    if (!something) { return; }
+    doThis();
+    doThat();
+  }
   ```
 
 [↑ scrollTo('#table-of-contents')](#table-of-contents)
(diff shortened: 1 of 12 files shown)


packages/eslint-plugin-shopify/docs/rules/prefer-class-properties.md
```diff
@@ -0,0 +1,50 @@
...
+```js
+class Foo {
+  constructor(baz) {
+    this.baz = baz;
+    this.qux = this.initialize();
+  }
+}
+```
+
+The following are considered wranings when setting the argument to `'never'`:
```
GoodForOneFare: Typo: `wranings`.

packages/eslint-plugin-shopify/docs/rules/prefer-class-properties.md
```diff
@@ -0,0 +1,50 @@
...
+```js
+class Foo {
+  baz = 123;
+  qux = something();
+}
+```
+
+## When Not To Use It
+
+If you do not wish to enforce the use or disuse of Flow for all files, then you can safely disable this rule.
```
GoodForOneFare: Copy+paste artifact?

packages/eslint-plugin-shopify/lib/config/esnext.js
```diff
@@ -26,6 +26,9 @@ module.exports = {
   rules: merge(
     require('./rules/ecmascript-6'),
     require('./rules/babel'),
-    {'no-param-reassign': 0} // because of default params
+    {
+      'no-param-reassign': 'off', // because of default params
```
GoodForOneFare: Sucks that this doesn't give a more granular config :/

README.md
```diff
@@ -1786,9 +1807,36 @@ npm run lint
   }
   ```
 
+- [12.11](#12.11) <a name="12.11"></a> When declaring constant instance members in a constructor, prefer class property syntax.
+
+  ESLint rule: [`prefer-class-properties`](packages/eslint-plugin-shopify/docs/prefer-class-properties.md)
+
+  > Why? This often removes the need for an explicit constructor and clearly indicates "setup" members for a class. It also matches how type declarations for instance properties are done in Flow.
```
GoodForOneFare: Really picky, but `It also matches how...` sounds clunky in my head.  I'd prefer something that puts the important part of the sentence first: `It also matches Flow's type declaration and instance property syntax`.

I'm okay with this staying as is, though.

packages/eslint-plugin-shopify/lib/rules/prefer-class-properties.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    return true;
+  }
+
+  function checkConstructorThisAssignment(node) {
+    if (isSimpleLiteral(node.right) && isStaticMemberExpression(node.left)) {
+      context.report({
+        node: node,
+        message: 'BAD!',
```
GoodForOneFare: Is there any chance of eventually making this message more descriptive?

lemonmade: lol, whoops, old commit that didn't get cleaned up :S

README.md
lemonmade: You better watch out, keep providing good suggestions and I will make you write all my docs :P

packages/eslint-plugin-shopify/lib/rules/prefer-class-properties.js
```diff
@@ -0,0 +1,83 @@
...
+      });
+    }
+  }
+
+  function checkClassDeclaration(node) {
+    if (applyAlways) {
```
GoodForOneFare: Maybe I just need some coffee beans, but I'd much prefer this to be calling methods with meaningful names.

```
checkClassDeclaration(node)
   if (applyAlways)
    var constructor = findConstructor(node)
   findMemberAssignments(constructor).forEach(... check ...)
  else
    findClassProperties(node).forEach(... BAD ...)
```

Even this has two distinct responsibilities, but I don't have to sift through all of those `===`s to get to the important stuff.

GoodForOneFare:
The usual pedantry from me, but  👍.  The only blocker is a typo and the `When Not To Use It` section.

packages/eslint-plugin-shopify/lib/rules/prefer-class-properties.js
lemonmade: Yeah, I was thinking something along these lines too and was too lazy. Good suggestion 👍

PR #45, opened 2016-04-19
Add useful plugins

### LINKED ISSUE #38: Look into adding/ integrating other ESLint plugins
Some good ones that might be relevant to us:

- [x] https://www.npmjs.com/package/eslint-plugin-chai-expect
- [x] https://www.npmjs.com/package/eslint-plugin-promise
- [x] https://www.npmjs.com/package/eslint-plugin-mocha
- [x] https://www.npmjs.com/package/eslint-plugin-jsx-a11y
- [x] https://www.npmjs.com/package/eslint-plugin-wpcalypso
- [x] https://www.npmjs.com/package/eslint-plugin-ava
- [x] https://www.npmjs.com/package/eslint-plugin-node
- [x] https://www.npmjs.com/package/eslint-plugin-flowtype

lemonmade:
Closes #38. This PR adds a bunch of plugins that address tools/ features of JS that we use at Shopify. It also adds specific configurations for tools that can be used to augment the "core" configurations (es5/ esnext/ react). These include:

- Plugin/ config for mocha/ chai/ sinon.
- Plugin/ config for Ava
- Plugin/ config for flow

I also cleaned up how the core configurations were structured. This includes removing the `node` env from all core configs, and moving it into a new, dedicated config that can be used in conjunction with the core ones. Also updated all documentation and projects to work correctly with the new configs. This actually caught an exclusive test we were doing in `shopify-codemod`, so it's already paying dividends!

I don't think the rule selections are particularly controversial, but let me know if there's anything you think is a bad choice.

cc/ @GoodForOneFare

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,21 +23,21 @@
   "author": "Chris Sauve <chrismsauve@gmail.com>",
   "license": "MIT",
   "devDependencies": {
-    "babel-cli": "^6.6.4",
-    "babel-core": "^6.7.0",
+    "babel-cli": "^6.7.5",
+    "babel-core": "^6.7.6",
     "babel-preset-shopify": "file:./packages/babel-preset-shopify",
     "chai": "^3.5.0",
     "coveralls": "^2.11.8",
+    "eslint": "^2.8.0",
     "eslint-plugin-shopify": "file:./packages/eslint-plugin-shopify",
-    "eslint": "^2.5.1",
     "isparta": "^4.0.0",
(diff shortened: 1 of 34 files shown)


GoodForOneFare:
Overall, LGTM 👍 It'd be good to get a second set of eyeballs on stuff that I'm less familiar with (flow, flowtype, ava, jsx-a11y).

lemonmade:
I think we will ship as-is and people using those configs can help us fine-tune them.

PR #47, opened 2016-04-21
Make esify an installable binary

lemonmade:
This adds `esify` as an actual project into the JS repo, and moves things around so that it can work by being installed globally and running `esify path/to/file.coffee` from the root of the Shopify repo.

cc/ @GoodForOneFare @bouk

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,4 @@ coverage
 .DS_STORE
 npm-debug.log
 
-# generator-eslint-shopify
 packages/generator-eslint-shopify/generators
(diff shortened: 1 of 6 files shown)


bouk:
We probably shouldn't commit the compiled code

lemonmade:
Yeah, I rolled back the compiling stuff because the other transforms we use need babel to work properly anyways :(

GoodForOneFare:
Does this belong in here?  It seems like we'll just delete it once the conversion is complete.

GoodForOneFare:
Code LGTM, and 🎩 went well.

Note that I couldn't lint this; if I roll back to before @d0ab34203de7fc25aa0bd476548de130e7df1faa, things work again.

lemonmade:
@GoodForOneFare I think it's fine in here, simplifies how many repos we have and we can just delete it once it's no longer useful. I think the linting issues will probably go away with a fresh `script/setup`.

PR #49, opened 2016-04-25
Add no-fully-static-classes rule

lemonmade:
This PR adds a rule to prevent the use of classes with only static members. I see this a bunch, particularly from people that aren't really used to JavaScript, and it drives me completely crazy (obviously, an object literal would be better since it avoids the one level of extra object creation).

cc/ @GoodForOneFare @bouk


bouk:
They should be object *or* for good ES2015 style just separately exported functions/constants in a module

lemonmade:
Good point, I'll add that to the error message.

PR #54, opened 2016-05-02
Strip unnecessary parenthesis from template strings

GoodForOneFare:
Decaf wraps *every* template string with parenthesis.  There's nowhere in the Shopify codebase that requires this.

This transform relies on odd behaviour to strip out parenthesis; I'm guessing that recast sees a new property on the node and this forces a (good) destructive change.  There's an unsupported `Node#parenthesizedExpression` property that *may* make this less haphazard, but I couldn't get it to work any more effectively than this solution.

/cc @lemonmade

### DIFF
--- a/packages/esify/index.js
+++ b/packages/esify/index.js
@@ -17,6 +17,7 @@ var TRANSFORMS = [
   {path: 'shopify-codemod/transforms/global-reference-to-import'},
   {path: 'js-codemod/transforms/arrow-function'},
   {path: 'js-codemod/transforms/template-literals'},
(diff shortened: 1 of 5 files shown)


lemonmade:
Saw this in the `js-codemod` repo, maybe it's a more appropriate way to force reprinting? https://github.com/cpojer/js-codemod/blob/b627c7faaacbe3cbe48050aea70ebc9aa4684c9b/transforms/trailing-commas.js#L25-L27

lemonmade:
Code LGTM, can you add the sanity check for `if ((foo === bar) && (baz === qux)) {}`?

PR #61, opened 2016-05-04
Add CoffeeScript soak transform

lemonmade:
This PR adds a transform to make the output of CoffeeScript's `?` operator a lot cleaner. The output of that operator quickly becomes a complex, nested ternary operator, so this will be a big improvement.

cc/ @GoodForOneFare @bouk

### DIFF
--- a/packages/shopify-codemod/test/fixtures/coffeescript-soak-to-condition/assignment.output.js
+++ b/packages/shopify-codemod/test/fixtures/coffeescript-soak-to-condition/assignment.output.js
@@ -1,6 +1,5 @@
-if (typeof bar.baz === 'function') {
-  foo = bar.baz();
-}
+// foo; foo = bar.baz?()
+foo = (typeof bar.baz === 'function' ? bar.baz() : undefined);
 
 // foo += bar.baz?() || some.default() for bar in bars
 for (let bar of bars) {
@@ -9,3 +8,8 @@ for (let bar of bars) {
 
 // (foo = bar?.baz) ->
 function aFunc(foo = (bar != null ? bar.baz : undefined)) {}
+
+// if foo = baz?.qux
+if(foo = (bar = (baz != null ? baz.qux : undefined))) {
+
+}
--- a/packages/shopify-codemod/test/fixtures/coffeescript-soak-to-condition/assignment.output.js
+++ b/packages/shopify-codemod/test/fixtures/coffeescript-soak-to-condition/assignment.output.js
@@ -1,16 +1,11 @@
-if (typeof bar === 'function' && bar() != null && bar().baz.qux != null) {
-  foo = bar().baz.qux.buzz;
+if (typeof bar.baz === 'function') {
+  foo = bar.baz();
 }
 
-// foo += bar?()?.baz.qux?.buzz || some.default() for bar in bars
+// foo += bar.baz?() || some.default() for bar in bars
 for (let bar of bars) {
-  foo += ((((typeof bar === 'function' && bar() != null && bar().baz.qux != null ? bar().baz.qux.buzz : undefined)))) || some.default();
+  foo += (((typeof bar.baz === 'function' ? bar.baz() : undefined))) || some.default();
 }
 
 // (foo = bar?.baz) ->
 function aFunc(foo = (bar != null ? bar.baz : undefined)) {}
-
-// foo = bar?()?.baz.qux?.buzz
-// bar = foo.something()
-var foo = (typeof bar === 'function' && bar() != null && bar().baz.qux != null ? bar().baz.qux.buzz : undefined),
-  bar = foo.somethingElse();
--- a/packages/shopify-codemod/test/fixtures/coffeescript-soak-to-condition/assignment.output.js
+++ b/packages/shopify-codemod/test/fixtures/coffeescript-soak-to-condition/assignment.output.js
@@ -1,0 +1,11 @@
+if (typeof bar === 'function' && bar() != null && bar().baz.qux != null) {
+  foo = bar().baz.qux.buzz;
+}
+
+// foo += bar?()?.baz.qux?.buzz || some.default() for bar in bars
+for (let bar of bars) {
+  foo += ((((typeof bar === 'function' && bar() != null && bar().baz.qux != null ? bar().baz.qux.buzz : undefined)))) || some.default();
+}
+
+// (foo = bar?.baz) ->
+function aFunc(foo = (bar != null ? bar.baz : undefined)) {}
--- a/packages/shopify-codemod/test/fixtures/coffeescript-soak-to-condition/assignment.output.js
+++ b/packages/shopify-codemod/test/fixtures/coffeescript-soak-to-condition/assignment.output.js
@@ -9,3 +9,8 @@ for (let bar of bars) {
 
 // (foo = bar?.baz) ->
 function aFunc(foo = (bar != null ? bar.baz : undefined)) {}
+
+// foo = bar?()?.baz.qux?.buzz
+// bar = foo.something()
+var foo = (typeof bar === 'function' && bar() != null && bar().baz.qux != null ? bar().baz.qux.buzz : undefined),
+  bar = foo.somethingElse();
--- a/packages/esify/index.js
+++ b/packages/esify/index.js
@@ -5,12 +5,15 @@ var jscodeshift = require('jscodeshift');
 require('babel-register')({ignore: false});
 
 var TRANSFORMS = [
+  {path: 'shopify-codemod/transforms/coffeescript-soak-to-condition'},
   {path: 'shopify-codemod/transforms/constant-function-expression-to-statement'},
   {path: 'shopify-codemod/transforms/ternary-statement-to-if-statement'},
   {path: 'shopify-codemod/transforms/mocha-context-to-closure', test: true},
   {path: 'shopify-codemod/transforms/mocha-context-to-global-reference', test: true},
   {path: 'shopify-codemod/transforms/coffeescript-range-output-to-helper'},
   {path: 'shopify-codemod/transforms/remove-useless-return-from-test', test: true},
+  {path: 'shopify-codemod/transforms/remove-addeventlistener-returns'},
+  {path: 'shopify-codemod/transforms/remove-empty-returns'},
   {path: 'shopify-codemod/transforms/conditional-assign-to-if-statement'},
   {path: 'shopify-codemod/transforms/function-to-arrow'},
   {path: 'shopify-codemod/transforms/global-assignment-to-default-export', test: false},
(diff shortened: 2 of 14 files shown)


bouk:
Is this something that could be put into decaf itself?

packages/shopify-codemod/test/fixtures/coffeescript-soak-to-condition/assignment.output.js
```diff
@@ -0,0 +1,16 @@
+if (typeof bar === 'function' && bar() != null && bar().baz.qux != null) {
```
bouk: This code is not the same as the previous one, as `bar()` is executed multiple times

lemonmade:
It's tricky to put into decaf because a bunch of these rely on making major changes, like switching variable declarations for if statements. When I tried it in decaf it was too convoluted to do anything other than a few basic ones I added a while ago.

packages/shopify-codemod/test/fixtures/coffeescript-soak-to-condition/assignment.output.js
lemonmade: You're right, I did it this way to try and keep things simple/ without extra intermediary variables. I looked through the admin and it doesn't look like there's anywhere that we do this that would cause issues (that is, no conditionally-called methods that have side effects), it will just be a little less efficient. Not sure if it's best to add the complexity, or to just bail out if a case like this is found.

GoodForOneFare:
Saw some errors during 🎩:
```
undefined does not match field "left": Expression of type BinaryExpression
    app/assets/javascripts/admin/initializers/handlers.coffee
    app/assets/javascripts/admin/lib/deduped_xhr.coffee
    app/assets/javascripts/admin/lib/embedded_app_hosts/admin_embedded_app_host.coffee
    app/assets/javascripts/admin/modules/address_operations.coffee
    app/assets/javascripts/admin/modules/form_with_poller.coffee
    app/assets/javascripts/admin/modules/order_address_editor.coffee
    app/assets/javascripts/admin/modules/payments/activation_forms/payments_activation_form.coffee
    app/assets/javascripts/admin/modules/products/product_importer.coffee
    app/assets/javascripts/admin/modules/rte/insert_image_modal.coffee
    app/assets/javascripts/admin/modules/rte/modal.coffee
    app/assets/javascripts/admin/modules/rte/rte.coffee
    app/assets/javascripts/admin/modules/spreadsheet_navigator.coffee
    app/assets/javascripts/admin/modules/template_editor.coffee
    app/assets/javascripts/checkout_next/shared/province_selector.coffee
    app/assets/javascripts/design_mode_next/outlines.coffee
    app/assets/javascripts/design_mode_next/overlay.coffee
    app/assets/javascripts/external/analytics.coffee
    app/assets/javascripts/partners/modules/country_choice.coffee
    app/assets/javascripts/sello/components/form_fields.coffee
    app/assets/javascripts/sello/components/ui_tip.coffee
    app/assets/javascripts/sello/form_handlers/account_update.coffee
    app/assets/javascripts/sello/form_handlers/payments/payments_activation_form.coffee
    app/assets/javascripts/sello/form_handlers/sign_up.coffee
    app/assets/javascripts/signup/details_form.coffee
    app/assets/javascripts/theme_editor_next/settings/image.coffee
    app/assets/javascripts/theme_editor_next/sidebar/block_list.coffee
    app/assets/javascripts/theme_editor/pickers/image_picker.coffee
    app/assets/javascripts/turbo/bar.coffee
    app/assets/javascripts/turbo/toxiproxy.coffee
Unexpected token
    app/assets/javascripts/admin/lib/autocomplete_v2/autocomplete_v2.coffee (112:8)
    app/assets/javascripts/admin/lib/resource_browser/resource_browser.coffee (115:8)
    app/assets/javascripts/admin/modules/timeline/attachment_drag_helper.coffee (49:8)
```

GoodForOneFare:
In `admin/lib/autocomplete/abstract-popover-autocomplete.js`, the JS translation is missing a `slice` call (it's fine if I esify without this transform).

CS:
```
  renderSuggestionsList: (suggestions) ->
    @suggestions = suggestions?[0...Shopify.AUTOCOMPLETE_MAX_SIZE]
```

JS:
```
  renderSuggestionsList(suggestions) {
    if (suggestions != null) {
      this.suggestions = suggestions(0, Shopify.AUTOCOMPLETE_MAX_SIZE);
    }
```

GoodForOneFare:
In `admin/lib/error-tracking.js`, there's no `isImmediatePropagationStopped()` call.

CS
```
  submit: (event) =>
    Shopify.Form._dirtyForm = false
    retValue = @opts.submit?.apply(this, arguments)
    @startLoading() unless event?.isImmediatePropagationStopped()
    return retValue
```

JS
```
  submit(event) {
    let ref;
    Shopify.Form._dirtyForm = false;

    if (this.opts.submit != null) {
      var retValue = this.opts.submit(this, arguments);
    }

    if (!((((event != null ? event() : undefined))))) {
      this.startLoading();
    }

    return retValue;
  }
```

lemonmade:
@GoodForOneFare can you take a look again? Pushed some changes to fix those issues and make these checks inside binary expressions nicer.

GoodForOneFare:
May be missing explicit clearing of properties in transforms like this from `admin/lib/autocomplete/abstract-popover-autocomplete.js`:

After decaf
```
  renderSuggestionsList(suggestions) {
    this.suggestions = typeof suggestions !== 'undefined' && suggestions !== null ? suggestions.slice(0, Shopify.AUTOCOMPLETE_MAX_SIZE) : void 0;
```

After Esify
```
 renderSuggestionsList(suggestions) {
    if (suggestions != null) {
      this.suggestions = suggestions.slice(0, Shopify.AUTOCOMPLETE_MAX_SIZE);
    }
```

GoodForOneFare:
Oddly nested `if` in `admin/lib/autocomplete_v2/autocomplete-v2.js`:
```
   if (if (this.input != null) {
     ref = this.input.trim();
   }) {
```

GoodForOneFare:
Might be worth raising an issue to create an issue to convert `typeof foo === 'function'` to `_.isFunction(foo)`.

GoodForOneFare:
I think an `asset.fromJSON(json)` call is accidentally stripped from `admin/modules/template-editor.js`:

Before soak transform:
```
          if (json = (ref = request.responseJSON) != null ? ref.asset : void 0) {
            asset.fromJSON(json);
          }
```

After soak transform:
```
          if (request.responseJSON != null) {
            json = request.responseJSON.asset;
          }
```

GoodForOneFare:
`ui-popover.js` contains an unsoaked `ref`:

Without soak transform:
```
    isActive() {
      let ref;
      if (popover === (((ref = activeCache.popover) != null ? ref.base : void 0))) {
```

After soak transform:
```
    isActive() {
      let ref;
      return activeCache.popover != null && activeCache.popover.base === (((ref = activeCache.popover) != null ? ref.base : void 0));
```

packages/shopify-codemod/test/fixtures/coffeescript-soak-to-condition/assignment.output.js
GoodForOneFare: I've spotted a couple of places where this is a performance concern (parsing JSON responses, triggering reflows).  My PR to simplify some CoffeeScript files will take care of this.

GoodForOneFare:
Another disappeared line in `ui-popover.js` (no `return` statement).
```diff
       let base;
       let forId;

-      if (forId = typeof (base = event.target).getAttribute === "function" ? base.getAttribute("for") : void 0) {
-        const fromLabel = true;
-        return $((`#${forId}`));
-      } else {
-        return $(event.target);
+      if (typeof event.target.getAttribute === 'function') {
+        forId = event.target.getAttribute("for");
       }
```

GoodForOneFare:
Oddly nested if in `sello/components/form-fields.js`:
```diff
   let ref1;
   let ref;
-  const disabled = (ref = ((ref1 = containerForField(field)) != null ? ref1.classList.contains(STATES.CONTAINER.DISABLED) : void 0)) != null ? ref : field.disabled;
+  const disabled = (if (containerForField(field) != null) {
+    ref = containerForField(field).classList.contains(STATES.CONTAINER.DISABLED);
+  } ? ref : field.disabled);
```

packages/shopify-codemod/test/fixtures/coffeescript-soak-to-condition/assignment.output.js
bouk: I feel very strongly that we shouldn't change the effect of code, we could introduce some really insidious bugs this way

lemonmade: Yeah, gave it some thought and I think that bailing out when we encounter this scenario is the best choice.

lemonmade:
@GoodForOneFare can you take one more look?

GoodForOneFare:
I probably won't get this finished today, but this works perfectly on all `shopify/shopify` code, and the output looks fantastic.

GoodForOneFare:
`admin/lib/autocomplete_v2/autocomplete-v2.js`
Just sanity checking here.  The `ref` is being intentionally retained so that its result can be returned without calling the function twice?

```diff
-    if ((ref = ((ref1 = this.input) != null ? ref1.trim() : void 0)) != null) {
+    if ((ref = ((this.input != null ? this.input.trim() : undefined))) != null) {
        return ref;
```

GoodForOneFare:
`admin/modules/template-editor.js`
Don't think this is a concern, but it's a theoretical change in behaviour.  No action required; haven't seen this affect anything else.
```diff
-      this.newAsset.isDuplicate = !!this.assets[`assets/${(ref = this.newAsset.file) != null ? ref.key : void 0}`];
+      this.newAsset.isDuplicate = !!this.assets[this.newAsset.file != null && `assets/${this.newAsset.file.key}`];
```

GoodForOneFare:
Stuff that should be tracked but definitely doesn't have to be addressed by this PR:
- Clean up excess parens
- Remove unused `ref`/`ref1` declarations
- Remove unnecessary ternarys, e.g., the `: undefined` part of these isn't useful
```js
 return ((((this.suggestions != null ? this.suggestions[index] : undefined)))) === this.input;
```
```js
  if (!((((event != null ? event.isImmediatePropagationStopped() : undefined))))) {
```

GoodForOneFare:
Everything checks out for me 👍  This is a heroic transform 🚢 🚢 🚢

lemonmade:
@GoodForOneFare thanks for the thorough review. I think our linting config will make it easy to pull those variables out, the transform to remove them would be quite tricky. For your last example, the first `: undefined` is actually necessary, since `this.input` could be `undefined`. I added a better transform for unary ! expressions, thanks for catching that!

GoodForOneFare:
Thanks for taking care of the ternaries 🐑  Linting out the remaining refs sounds good to me.

PR #62, opened 2016-05-05
Add constructor-literal-assignment-to-class-property transform

lemonmade:
This PR adds a transform that is basically just for enforcing the `prefer-class-property` ESLint rule we added. It moves any context assignments in the constructor that are not dynamic (i.e., that don't depend on `this` and that don't depend on any bindings only available in the constructor) to be class properties.

cc/ @GoodForOneFare @bouk @fandy


packages/shopify-codemod/test/fixtures/constructor-literal-assignment-to-class-property/complex-literals.output.js
```diff
@@ -0,0 +1,11 @@
+class Foo {
+  foo = {bar: 12, baz: true};
+  bar = {[baz]: 'something'};
```
bouk: does this make the eslint rule fail if we don't put it as a class property? because it's pretty confusing to have to intersperse the contructor and these lines to be able to make sense of it (like, there's no easy way to see where baz came from)

lemonmade: The ESLint rule actually doesn't fail on these, because I don't think ESLint can do the same scope stuff that we can do with jscodeshift. The ESLint rule should only fail for properties that are either string/ boolean/ numeric/ regexp literals, objects/ arrays that are completely simple literals, or method calls on one of those first two.

lemonmade:
We're going to turn off the ESLint rule that prompted this transform, so reviewing this isn't super important. I think I will still merge so that anyone who wants it can have it, though.

PR #73, opened 2016-05-10
Transform sinon assertions to use methods that generate meaningful error messages

GoodForOneFare:
* Does some cleanup to get consistently searchable assertions
* Turns any inline `foo.called`-like assertions into `assert.called(foo)`-like assertions

Note: 2 cases won't be cleaned up because they don't map to a sinon assertion.  They'll be handled in another PR.
- `assert.isFalse(foo.threw())`
- `assert.isFalse(foo.calledWithExactly(bar))`

/cc @lemonmade, @fandy

### DIFF
--- a/packages/esify/index.js
+++ b/packages/esify/index.js
@@ -21,6 +21,18 @@ var TRANSFORMS = [
   {path: 'js-codemod/transforms/object-shorthand'},
   {path: 'js-codemod/transforms/no-vars'},
   {path: 'js-codemod/transforms/unquote-properties'},
+  // Order is significant for these initial assert transforms; think carefully before reordering.
+  {path: 'shopify-codemod/transforms/assert/assert-false-to-assert-fail', test: true},
+  {path: 'shopify-codemod/transforms/assert/assert-to-assert-ok', test: true},
+  {path: 'shopify-codemod/transforms/assert/negated-assert-ok-to-assert-not-ok', test: true},
+  {path: 'shopify-codemod/transforms/assert/move-literals-to-expected-argument', test: true},
+  {path: 'shopify-codemod/transforms/assert/equality-comparisons-to-assertions', test: true},
+  // These transforms can be executed in any order.
+  {path: 'shopify-codemod/transforms/assert/called-equals-boolean-to-assert-called', test: true},
+  {path: 'shopify-codemod/transforms/assert/call-count-equals-to-assert-called', test: true},
+  {path: 'shopify-codemod/transforms/assert/called-method-to-assert-called', test: true},
+  {path: 'shopify-codemod/transforms/assert/called-with-methods-to-assert-called-with', test: true},
+  {path: 'shopify-codemod/transforms/assert/falsy-called-method-to-assert-not-called', test: true},
 ];
 
 var OPTIONS = loadOptions();
--- a/packages/shopify-codemod/test/fixtures/assert/falsy-called-method-to-assert-not-called/basic.output.js
+++ b/packages/shopify-codemod/test/fixtures/assert/falsy-called-method-to-assert-not-called/basic.output.js
@@ -1,0 +1,2 @@
+assert.neverCalledWith(foo, bar, qux);
+assert.neverCalledWithMatch(foo, bar);
--- a/packages/shopify-codemod/test/fixtures/assert/move-literals-to-expected-argument/basic.output.js
+++ b/packages/shopify-codemod/test/fixtures/assert/move-literals-to-expected-argument/basic.output.js
@@ -1,0 +1,13 @@
+assert.equal(foo, 5, 'error message');
+assert.equal(bar, 'foo');
+assert.equal(this.foo.bar(), 'foo');
+assert.equal(!bar, 'foo');
+assert.equal(foo, null);
+assert.equal(bar, `foo`);
+
+assert.equal(foo, []);
+assert.deepEqual(foo, {}, 'error message');
+assert.notDeepEqual(foo, {});
+assert.notEqual(foo, 1);
+assert.notStrictEqual(foo, 2);
+assert.strictEqual(foo, 'foo');
--- a/packages/shopify-codemod/transforms/assert/assert-to-assert-ok.js
+++ b/packages/shopify-codemod/transforms/assert/assert-to-assert-ok.js
@@ -1,0 +1,15 @@
+import {createAssertion} from './utils';
+
+export default function assertToAssertOk({source}, {jscodeshift: j}, {printOptions = {}}) {
+  return j(source)
+    .find(j.CallExpression, {
+      callee: {
+        type: j.Identifier.name,
+        name: 'assert',
+      },
+    })
+    .replaceWith(({node}) =>
+      createAssertion('ok', node.arguments)
+    )
+    .toSource(printOptions);
+}
--- a/packages/shopify-codemod/transforms/assert/call-count-equals-to-assert-called.js
+++ b/packages/shopify-codemod/transforms/assert/call-count-equals-to-assert-called.js
@@ -1,0 +1,46 @@
+import {createAssertion, isAssert, isMemberCall} from './utils';
+
+export default function callCountEqualsToAssertCalled({source}, {jscodeshift: j}, {printOptions = {}}) {
+  function intToCallMethodName(callCount) {
+    return {
+      0: 'notCalled',
+      1: 'calledOnce',
+      2: 'calledTwice',
+      3: 'calledThrice',
+    }[callCount];
+  }
+
+  const sourceAST = j(source);
+  sourceAST
+    .find(j.CallExpression, {
+      callee: isAssert('equal', 'strictEqual', 'deepEqual'),
+      arguments: {
+        0: isMemberCall('callCount'),
+        1: (node) => node.value >= 0 && node.value <= 3,
+      },
+    })
+    .replaceWith(({node: {arguments: [actualArg, expectedArg]}}) => {
+      const assertionName = intToCallMethodName(expectedArg.value);
+
+      return j.callExpression(
+        j.memberExpression(j.identifier('assert'), j.identifier(assertionName), false),
+        [actualArg.object]
+      );
+    });
+
+  return sourceAST
+    .find(j.CallExpression, {
+      callee: isAssert('equal', 'strictEqual', 'deepEqual'),
+      arguments: {
+        0: isMemberCall('callCount'),
+        1: (node) =>
+          j.MemberExpression.check(node) ||
+          j.Identifier.check(node) ||
+          j.Literal.check(node) && isFinite(node.value),
+      },
+    })
+    .replaceWith(({node: {arguments: [actualArg, expectedArg]}}) =>
+      createAssertion('callCount', [actualArg.object, expectedArg])
+    )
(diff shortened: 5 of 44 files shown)


packages/esify/index.js
```diff
@@ -20,6 +20,16 @@ var TRANSFORMS = [
   {path: 'shopify-codemod/transforms/function-to-arrow'},
   {path: 'shopify-codemod/transforms/global-assignment-to-default-export', test: false},
   {path: 'shopify-codemod/transforms/global-reference-to-import'},
+  {path: 'shopify-codemod/transforms/assert/assert-to-assert-ok', test: true},
+  {path: 'shopify-codemod/transforms/assert/negated-assert-ok-to-assert-not-ok', test: true},
```
lemonmade: Do these two assertions need to come first in order for the rest to work?

packages/shopify-codemod/test/fixtures/assert/falsy-called-method-to-assert-not-called/basic.output.js
```diff
@@ -0,0 +1,2 @@
+assert.neverCalledWith(foo, bar, qux);
```
lemonmade: Sorry if this is a silly comment once I have more context, but don't want to forget: will this properly deal with something like `assert.notOK(foo.calledTwice)`?

packages/shopify-codemod/test/fixtures/assert/move-literals-to-expected-argument/basic.output.js
```diff
@@ -0,0 +1,11 @@
+assert.equal(foo, 5, 'error message');
```
lemonmade: Can you add an issue to have a linting rule to verify this?

packages/shopify-codemod/test/fixtures/assert/move-literals-to-expected-argument/basic.output.js
```diff
@@ -0,0 +1,11 @@
...
+assert.equal(bar, 'foo');
+assert.equal(this.foo.bar(), 'foo');
+assert.equal(!bar, 'foo');
+assert.equal(foo, null);
```
lemonmade: Does this handle them being used in the proper order, with an error message at the end? `assert.equal(foo, bar, 'error message')` should stay the same, just want to make sure it does.

packages/shopify-codemod/test/fixtures/assert/threw-method-to-assert-threw/basic.output.js
```diff
@@ -0,0 +1,2 @@
+assert.threw(foo);
+assert.threw(this.foo.bar);
```
lemonmade: Does this work with an argument passed to `threw`?

packages/shopify-codemod/transforms/assert/assert-to-assert-ok.js
```diff
@@ -0,0 +1,27 @@
...
+        name: 'assert',
+      },
+      arguments: {
+        0: {
+          type: (type) => [
+            j.BinaryExpression.name,
+            j.CallExpression.name,
+            j.Identifier.name,
+            j.MemberExpression.name,
+            j.UnaryExpression.name,
```
lemonmade: What types are we specifically trying to exclude here?

packages/shopify-codemod/transforms/assert/assert-to-assert-ok.js
```diff
@@ -0,0 +1,27 @@
...
+          ].indexOf(type) > -1,
```
lemonmade: Might be nicer to use a `Set` in scope instead of an array constructed each time.

packages/shopify-codemod/transforms/assert/call-count-equals-to-assert-called.js
```diff
@@ -0,0 +1,49 @@
+import {isAssert, isMemberCall} from './utils';
+
+export default function callCountEqualsToAssertCalled({source}, {jscodeshift: j}, {printOptions = {}}) {
+  function intToCallMethodName(callCount) {
+    return {
```
lemonmade: Might be nice to set this up in scope instead of declaring each time.

packages/shopify-codemod/transforms/assert/call-count-equals-to-assert-called.js
```diff
@@ -0,0 +1,49 @@
...
+    })
+    .replaceWith(({node: {arguments: [actualArg, expectedArg]}}) => {
+      const assertionName = intToCallMethodName(expectedArg.value);
+
+      return j.callExpression(
+        j.memberExpression(j.identifier('assert'), j.identifier(assertionName), false),
+        [actualArg.object]
+      );
+    })
+    .toSource(printOptions);
```
lemonmade: Do we have to do the `toSource` call? I think we can just store `j(source)` and then do two separate `.find().replaceWith()` calls on it, before finally doing the `toSource` at the end.

packages/shopify-codemod/transforms/assert/call-count-equals-to-assert-called.js
```diff
@@ -0,0 +1,49 @@
...
+      2: 'calledTwice',
+      3: 'calledThrice',
+    }[callCount];
+  }
+
+  const sourceWithSmallCallCountsRemoved = j(source)
+    .find(j.CallExpression, {
+      callee: isAssert('equal', 'strictEqual', 'deepEqual'),
+      arguments: {
+        0: isMemberCall('callCount'),
```
lemonmade: Utilities make this really easy to read 👍

packages/shopify-codemod/transforms/assert/call-count-equals-to-assert-called.js
```diff
@@ -0,0 +1,49 @@
...
+        1: (node) =>
+          j.MemberExpression.check(node) ||
+          j.Identifier.check(node) ||
+          j.Literal.check(node) && isFinite(node.value),
+      },
+    })
+    .replaceWith(({node: {arguments: [actualArg, expectedArg]}}) =>
+      j.callExpression(
+        j.memberExpression(j.identifier('assert'), j.identifier('callCount'), false),
```
lemonmade: Another utility for creating these assert method calls might be nice.

packages/shopify-codemod/transforms/assert/called-method-to-assert-called.js
```diff
@@ -0,0 +1,41 @@
...
+          j.identifier('assert'),
+          j.identifier(invertCalledAssertionName(callArg.property.name)),
+          false
+        ),
+        [callArg.object]
+      )
+    )
+    .toSource(printOptions);
+
+  return j(sourceWithNegatedCalledsRemoved)
```
lemonmade: Same comment re: the `toSource` call.

packages/shopify-codemod/transforms/assert/equality-comparisons-to-assertions.js
```diff
@@ -0,0 +1,23 @@
+import {isAssert} from './utils';
+
+export default function equalityComparisonsToAssertions({source}, {jscodeshift: j}, {printOptions = {}}) {
+  return j(source)
+    .find(j.CallExpression, {
+      callee: isAssert('isTrue', 'ok'),
+      arguments: {
+        0: {
+          type: j.BinaryExpression.name,
+          operator: (op) => ['===', '!=='].indexOf(op) > -1,
```
lemonmade: Set might be nicer here too.

packages/shopify-codemod/transforms/assert/move-literals-to-expected-argument.js
```diff
@@ -0,0 +1,27 @@
+import {isAssert} from './utils';
+
+export default function moveLiteralsToExpectedArgument({source}, {jscodeshift: j}, {printOptions = {}}) {
+  function isHardcodedValue(type) {
+    return [j.Literal.name, j.ObjectExpression.name].indexOf(type) > -1;
```
lemonmade: Set

lemonmade: Also, do we not care about arrays here?

lemonmade:
Minor comments. Overall these look absolutely fabulous, extremely easy to read and very useful 👍

packages/esify/index.js
GoodForOneFare: Indeed.   Also `move-literals-to-expected-argument' and `equality-comparisons-to-assertions`.

I was going to be lazy and just document the dependency in the README, but I'm open to doing something more bullet proof.

packages/shopify-codemod/test/fixtures/assert/move-literals-to-expected-argument/basic.output.js
GoodForOneFare: Ah, that's a good call: #74

packages/esify/index.js
lemonmade: I think we should add some comments in here, that will probably be the best way to prevent random changes in order.

packages/shopify-codemod/test/fixtures/assert/move-literals-to-expected-argument/basic.output.js
GoodForOneFare: It *should*.  I'll add a test to sanity check that.

packages/shopify-codemod/test/fixtures/assert/threw-method-to-assert-threw/basic.output.js
GoodForOneFare: No, and now that I look at it, it's horribly broken in other ways :/  I'm probably going to remove this from the PR, because the fix isn't trivial.

packages/shopify-codemod/transforms/assert/assert-to-assert-ok.js
GoodForOneFare: Fantastic question.  Short answer: it's trying to work around an issue that I should fix with another transform.

Long answer: It's  (verbosely) excluding `false` to avoid dodgy assertions like `assert(false, "something bad happened");` being affected by subsequent transforms.  Of course, there should be a `assert-false-to-assert-fail` transform 👍

packages/shopify-codemod/test/fixtures/assert/falsy-called-method-to-assert-not-called/basic.output.js
GoodForOneFare: The `called-method-to-assert-called` handles similar cases.  Nothing like `assert.notOK(foo.calledTwice)` appears in our code, so I didn't expend any effort on it.  

Which assertion would provide the most meaningful message for that scenario?  `assert.notEquals(foo.callCount, 2)`?

lemonmade: If it doesn't appear in our code, then don't worry about it (I think the example you put is probably the best translation).

GoodForOneFare:
@lemonmade All comments have been addressed, and `assert-false-to-assert-fail` now exists.  I'll handle `assert(foo.threw)` cleanup in a separate PR.

lemonmade:
👍

PR #76, opened 2016-05-11
Remove redundant returns from addEventListener

fandy:
Removes returns in `addEventListener` from the expression.

https://astexplorer.net/#/nRUFS7T4hj/8


packages/shopify-codemod/transforms/remove-redundant-returns-from-addeventlistener.js
```diff
@@ -0,0 +1,28 @@
+export default function transformer(file, api) {
```
GoodForOneFare: Please unpack `j`, `source`, etc using the same style as our other transforms, e.g.,:

`export default function functionToArrow({source}, {jscodeshift: j}, {printOptions = {}}) {`

GoodForOneFare: Please give the function a camel-case version of the file name.  A meaningful name will be easier to eyeball in stack traces.

packages/shopify-codemod/transforms/remove-redundant-returns-from-addeventlistener.js
```diff
@@ -0,0 +1,28 @@
...
+  const j = api.jscodeshift;
+  return j(file.source) // parse input to AST
```
GoodForOneFare: Don't need this comment.

fandy:
Updated with the latest commit.

packages/shopify-codemod/transforms/remove-redundant-returns-from-addeventlistener.js
```diff
@@ -0,0 +1,28 @@
...
+  return j(file.source)
+    .find(j.CallExpression, {
+      callee: {
+        type: 'MemberExpression',
+        property: {
+          name: 'addEventListener',
+        },
+      },
+    })
+    .forEach((nodes) => {
```
GoodForOneFare: `nodes` is a misleading name because an array isn't being passed in here, and it's actually a `NodePath`.

I'd recommend using destructuring to keep it concise. e.g., `.forEach(({node}) => {`

packages/shopify-codemod/transforms/remove-redundant-returns-from-addeventlistener.js
```diff
@@ -0,0 +1,28 @@
+export default function removeAddEventListener(file, api) {
```
GoodForOneFare: The transform fails if I give it this input:

```
document.addEventListener('input', event => {
  event.preventDefault();
  return;
}, true);
```

packages/shopify-codemod/transforms/remove-redundant-returns-from-addeventlistener.js
```diff
@@ -0,0 +1,28 @@
...
+          body.pop();
+        }
+        body[body.length - 1] = j.expressionStatement(body[body.length - 1].argument);
+      } else {
+        return nodes;
+      }
+      return body;
+    }
+  )
+  .toSource();
```
GoodForOneFare: Please pass in `printOptions` here.  This'll ensure that the generated code uses the correct style rules (whitespace, quoting, etc).

GoodForOneFare:
The astexplorer link is a nice touch, but there should be some unit tests before we merge this.  Be sure to include a test case to prove that meaningful returns are *not* affected.

Grab me or @lemonmade when you're ready to work on tests.  It's a simple setup, but a quick intro can't hurt :)

packages/shopify-codemod/transforms/remove-redundant-returns-from-addeventlistener.js
```diff
@@ -0,0 +1,28 @@
...
+  const j = api.jscodeshift;
+  return j(file.source)
+    .find(j.CallExpression, {
+      callee: {
+        type: 'MemberExpression',
+        property: {
+          name: 'addEventListener',
+        },
+      },
+    })
```
GoodForOneFare: If the `find` was more specific, the `forEach` wouldn't have to do so much logic.

```
      callee: ... same as before ...,
      arguments: {
        1: {
          type: j.ArrowFunctionExpression.name,
          body: {
            body: (lines) => j.ReturnStatement.check(lines[lines.length-1])
          }          
        }
      }
```

The pro of this is that it keeps selection / transforming in separate blocks of code.  Easy for people to comprehend.

The con of this is that it makes smarter transforms harder, but creating a separate transform for edge cases is usually the right way to go.

GoodForOneFare:
Oh, hey, this is a really nice first transform. Great job ramping up and sending out code for feedback so quickly.

packages/shopify-codemod/transforms/remove-redundant-returns-from-addeventlistener.js
lemonmade: (You can copy the arguments to the exported function/ what to do at the bottom here from the other transforms.

packages/shopify-codemod/transforms/remove-redundant-returns-from-addeventlistener.js
```diff
@@ -0,0 +1,28 @@
...
+    .forEach((nodes) => {
+      const functionExpression = nodes.value.arguments[1];
+      const {body} = functionExpression.body;
+      const lastStatement = body[body.length - 1];
+      if (lastStatement.type === 'ReturnStatement') {
+        if (lastStatement.argument === null) {
+          body.pop();
```
lemonmade: Probably want to return here to avoid the undefined expression issue Gord mentioned.

packages/shopify-codemod/transforms/remove-redundant-returns-from-addeventlistener.js
lemonmade: What are the thoughts on also having this transform remove any arguments to `return` statements nested in blocks within the function?

lemonmade:
@fandy nice job on the first transform! Make sure to ping myself and Gord when you are ready for review on any PRs.

packages/shopify-codemod/transforms/remove-redundant-returns-from-addeventlistener.js
GoodForOneFare: From snooping at esified code, removing `return` from nested statements would be useful.  But it's probably best to do it in a separate transform.

fandy:
Moved the selector logic to `find` and handles null arguments. Should be ready for a review. /cc @lemonmade @GoodForOneFare

fandy:
Added tests /cc @GoodForOneFare

packages/shopify-codemod/test/transforms/remove-addeventlistener-returns.js
```diff
@@ -0,0 +1,14 @@
...
+import removeAddEventListenerReturns from 'remove-addeventlistener-returns';
+
+describe('removeAddEventListenerReturns', () => {
+  it('removes redundant returns with arguments from addEventListener', () => {
+    expect(removeAddEventListenerReturns).to.transform('remove-addeventlistener-returns/basic');
+  });
+  it('removes redundant returns with no arguments', () => {
+    expect(removeAddEventListenerReturns).to.transform('remove-addeventlistener-returns/no-argument');
+  });
+  it('don\'t remove nested returns', () => {
```
GoodForOneFare: `don't` => `doesn't`

GoodForOneFare:
Just one typo, then I think this is good to go! 🚢

fandy:
Fixed typo

PR #83, opened 2016-05-12
Removes redundant empty returns

fandy:
- Removes useless returns from `FunctionDeclaration`, 
- Removes useless returns from `ArrowFunctionExpression`
- Removes useless returns from`FunctionExpression`.


fandy:
What do you think? @GoodForOneFare @lemonmade

packages/shopify-codemod/transforms/remove-empty-returns.js
```diff
@@ -0,0 +1,35 @@
+export default function removeEmptyReturns({source}, {jscodeshift: j}, {printOptions = {}}) {
+  const sourceAST = j(source);
+
+  sourceAST
+  .find(j.FunctionDeclaration, {
```
lemonmade: Looks like all of these should be indented one extra level.

lemonmade: Also, I think we should be combine this into a single transform that just checks that the type is one of the three expressions we care about.

packages/shopify-codemod/transforms/remove-empty-returns.js
```diff
@@ -0,0 +1,35 @@
...
+    body: {
+      body: (lines) => j.ReturnStatement.check(lines[lines.length - 1]),
+    },
+  })
+  .forEach(({node}) => {
```
lemonmade: I like extra destructuring for this:

```js
.forEach(({node: {body: {body}}}) => {})
```

packages/shopify-codemod/transforms/remove-empty-returns.js
```diff
@@ -0,0 +1,35 @@
...
+    node.body.body[node.body.body.length - 1] = null;
```
lemonmade: `body.pop()` is probably better here.

packages/shopify-codemod/transforms/remove-empty-returns.js
```diff
@@ -0,0 +1,35 @@
+export default function removeEmptyReturns({source}, {jscodeshift: j}, {printOptions = {}}) {
+  const sourceAST = j(source);
+
+  sourceAST
+  .find(j.FunctionDeclaration, {
+    body: {
+      body: (lines) => j.ReturnStatement.check(lines[lines.length - 1]),
```
lemonmade: Won't this kill any trailing `return` statements, even if they have an actual argument?

packages/shopify-codemod/test/fixtures/remove-empty-returns/function-declarations.input.js
```diff
@@ -0,0 +1,5 @@
+function whenFinished() {
```
lemonmade: Can you add sanity checks for 1) returns that have an argument (not changed), and 2) returns that are nested inside a block (not changed)

packages/shopify-codemod/test/transforms/remove-empty-returns.js
```diff
@@ -0,0 +1,14 @@
+import 'test-helper';
```
GoodForOneFare: For the test runner to pick up this file, it needs a `.test.js` suffix.  i.e., these tests aren't being executed by `npm test` :)

GoodForOneFare:
@fandy tests+code in the same commit.  If you write the tests first, you'll notice stuff like the test runner ignoring them.

fandy:
PR should be ready now. Also added `npm run setup` to the README since it can be confusing to new Shopifolk.

packages/shopify-codemod/transforms/remove-empty-returns.js
```diff
@@ -0,0 +1,12 @@
+export default function removeEmptyReturns({source}, {jscodeshift: j}, {printOptions = {}}) {
+  return j(source)
+    .find(j.Function, {
+      body: {
+        body: (lines) => j.ReturnStatement.check(lines[lines.length - 1]) && lines[lines.length - 1].argument === null,
```
GoodForOneFare: I think you can import `matchlast` from `utils` and use it here.

packages/shopify-codemod/test/fixtures/remove-empty-returns/function-declarations.input.js
GoodForOneFare: I don't see the sanity tests that Chris requested.

packages/shopify-codemod/test/fixtures/remove-empty-returns/function-declarations.output.js
```diff
@@ -0,0 +1,7 @@
+function whenFinished() {
+  unhighlight(target, startTime);
```
GoodForOneFare: My preference would be to have less real world information in the tests.  It's harder to grok when I have to sort out that `whenFinished` and `unhighlight` don't have any significance to the solution.

The project's other tests tend to use `foo`, `bar`, `qux`, and/or convey information about the test.  e.g.,
```js
function returnShouldBeRemoved() {
  foo();
  bar();

  function anotherRemovedReturn() {
    qux();
    return;
  }

  return;
}
```

package.json
```diff
@@ -38,6 +38,9 @@
     "proxyquire": "^1.7.4",
     "shelljs": "^0.6.0",
     "sinon": "^1.17.3",
-    "sinon-chai": "^2.8.0"
+    "sinon-chai": "^2.8.0",
+    "yeoman-assert": "^2.2.1",
```
GoodForOneFare: I think these are not necessary.  I'm able to do a fresh checkout and get everything running without these dependencies.

```
git clone git@github.com:Shopify/javascript.git
cd javascript
script/setup
npm test
cd packages/shopify-codemod
npm test
```

packages/shopify-codemod/README.md
```diff
@@ -5,9 +5,10 @@ This repository contains a collection of Codemods written with [JSCodeshift](htt
 ## Usage
 
 1. `npm install -g jscodeshift`
-1. `git clone https://github.com/shopify/javascript` or [download the zip file](https://github.com/shopify/javascript/archive/master.zip)
-1. `jscodeshift -t <codemod-script> <file>`
-1. Use the `-d` option for a dry-run and use `-p` to print the output for comparison.
+2. `git clone https://github.com/shopify/javascript` or [download the zip file](https://github.com/shopify/javascript/archive/master.zip)
+3. `npm install && npm run setup`
```
GoodForOneFare: The `setup` script already invokes `npm install`.

fandy: Is it better to have `jscodeshift` as a local dependency? The reason is similar to why Babel doesn't recommend being installed globally:

1. Different projects on the same machine can depend on different versions of Babel allowing you to update one at a time.
2. It means you do not have an implicit dependency on the environment you are working in. Making your project far more portable and easier to setup.

fandy:
Just pushed an update - it's looking much cleaner now :)

packages/shopify-codemod/README.md
lemonmade: It would be fine if people did install `jscodeshift` locally, It's just easier to show it work when installed globally (since the executable is part of the path).

packages/shopify-codemod/test/fixtures/remove-empty-returns/arrow-functions.input.js
```diff
@@ -0,0 +1,4 @@
+$document.on('click', 'a.disabled, a.btn-disabled', event => {
+  event.preventDefault();
+  return;
+});
```
lemonmade: I still don't see the sanity checks for nested blocks and for returns with arguments:

```js
function foo() {
  if (bar) { return; }
  doSomething();
  return;
}
```

```js
function foo() {
  return 'bar';
}
```

fandy: Hmm, thought I added that. Will do this after writing my Pulse.

fandy:
Updated with sanity checks:

- don't remove returns in nested blocks
- don't remove returns with arguments

lemonmade:
👍 looks great! Remember to squash and this is good to go 🐑

GoodForOneFare:
👍 Good stuff.  Thank you!

PR #85, opened 2016-05-16
removes redundant empty returns

fandy:
combined transforms and corrected test cases

updated readme instructions to reflect setup

uses matchLast util and cleaned up dependencies + test case naming conventions

added sanity checks


fandy:
Fix of https://github.com/Shopify/javascript/pull/83

PR #87, opened 2016-05-18
Prefer equivalent JavaScript file for multiple global-to-import matches

### LINKED ISSUE #59: global-reference-to-import doesn't ignore transformed files
Once a file is transformed, `global-reference-to-import` flags Coffee+JavaScript exports as conflicting definitions.

e.g.,
```
Error: Found multiple definitions for Shopify.ElementQueries:
admin/lib/element_queries.coffee:136:1:Shopify.ElementQueries =
admin/lib/element-queries.js:1:1:'expose Shopify.ElementQueries';
```

lemonmade:
Fixes #59. This PR improves `global-reference-to-import` to allow duplicate import sources in one case: when there are exactly two matches, and they are an equivalent JS/ CS pair (same directory, JS name is dash-cased). I also added another assertion to `chai-jscodeshift` to expect a thrown error when transforming a particular fixture.

### DIFF
--- a/packages/chai-jscodeshift/index.js
+++ b/packages/chai-jscodeshift/index.js
@@ -22,27 +22,38 @@ module.exports = function chaiJSCodeShift(options) {
     var Assertion = chaiOptions.Assertion;
     var assert = chaiOptions.assert;
 
-    function transform(fixture, transformOptions) {
-      var transformer = this._obj;
-
+    function transformFixtureWithOptions(transformer, fixture, transformOptions) {
       var inputPath = finalConfig.inputFixturePath(fixture, finalConfig.fixtureDirectory);
-      var outputPath = finalConfig.outputFixturePath(fixture, finalConfig.fixtureDirectory);
 
       var input = fs.readFileSync(inputPath, 'utf8');
-      var output = fs.readFileSync(outputPath, 'utf8').trim();
-      var transformed = transformer(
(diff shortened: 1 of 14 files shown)


bouk:
Maybe a better fix is to delete the file immediately afterwards, which we want to do anyways (probably?)

Also another issue with transforming multiple files at once is: if the file has underscores in it it those will get transformed to dashes, which means the file reference will become invalid.

lemonmade:
I'm not sure deleting right away is a great choice, since you'll most likely want to refer back and forth between the compiled and new file in order to make sure the transformations worked properly.

The second issue is definitely real, we had talked about it a little before but didn't come to any consensus. IMO, it's not a huge deal since the bundle should fail to compile, which will be a pretty good signal to go back and update any imports. We could also add a script to check for whether any existing imports are pointing at the converted file and update them, or update all CoffeeScript files to be dash case now, or just let JavaScript files be snake case. Thoughts?

bouk:
I think we should just rename everything to dash case right now, use this PR and also add `git rm --cached` any file we transform so we don't accidentally leave it in the repo.

bouk:
PR lgtm by the way

GoodForOneFare:
👍

lemonmade:
I'm gonna think on the best course of action a little more, don't want to do anything that will be super disruptive.

PR #88, opened 2016-05-19
Improve order of transforms

lemonmade:
This PR moves the transforms around a little bit to have the stylistic ones at the end, which will hopefully catch and standardize any additional nodes that were added in subsequent transforms.


bouk:
Maybe we should apply all the transforms in a loop until we reach a stable position :trollface: 

lgtm

PR #89, opened 2016-05-19
esify now accepts multiple files/globs

GoodForOneFare:
Gets esify to accept multiple input args, e.g., 
```esify app/assets/javascripts/sello/**/*.coffee test/sello-javascripts/**/*.coffee```

I found this very useful while troubleshooting our soak transform problems, but I'm okay with not merging this if it's not generally useful.

/cc @lemonmade @bouk

### DIFF
--- a/packages/esify/bin/esify
+++ b/packages/esify/bin/esify
@@ -5,8 +5,8 @@ var glob = require('glob');
 var path = require('path');
 var transform = require('..');
 
-var pattern = process.argv[2];
-var files = getFiles(pattern);
+var patterns = process.argv.slice(2);
+var files = getFiles(patterns);
 var errors = {};
 
 if (files.length === 0) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/esify/bin/esify [hunk 1]
```diff
@@ -31,13 +31,17 @@ if (errorKeys.length > 0) {
 }
 
 function getFiles(filePattern) {
```
lemonmade: Probably want this to be plural now.

packages/esify/bin/esify
```diff
@@ -31,13 +31,17 @@ if (errorKeys.length > 0) {
...
-  if (filePattern == null) {
-    throw new Error('Pass in a .coffee file, or glob pattern.');
+  if (filePattern == null || filePattern.length === 0) {
+    throw new Error('Pass in .coffee files, or glob patterns.');
   }
 
-  return glob
-    .sync(filePattern, {ignore: ['node_modules/**', 'vendor/**']})
-    .filter(function(filePath) { return filePath.match(/\.coffee$/); });
+  return patterns.reduce(function(result, filePattern) {
```
lemonmade: Use `filePatterns` instead.

lemonmade:
Minor comments, no reason not to add it if it's something that you've been using 👍

packages/esify/bin/esify
GoodForOneFare: Ugh.  Fixed.  Thank you :)

PR #90, opened 2016-05-20
Minor improvements from a few more translation examples

lemonmade:
This PR does a few small things, I can unbundle them if it will make it easier to review:

- Adds `global-identifier-to-import` to turn globals (lodash, jQuery, etc) into import statements.
- Made `remove-useless-return-from-test` also remove the IIFE wrapper from the return argument, if present. This was causing some issues with the transforms that depend on scope to do things, like the mocha ones. The bare IIFE also doesn't really make much sense, it's mostly just used for CoffeeScript's everything-is-an-expression madness.
- Rejigged the order a little bit more.
- Made some small updates to the utilities and a few other files.

cc/ @GoodForOneFare @bouk @fandy

### DIFF
--- a/packages/esify/index.js
+++ b/packages/esify/index.js
@@ -37,6 +37,7 @@ var TRANSFORMS = [
   {path: 'shopify-codemod/transforms/strip-template-literal-parenthesis'},
   {path: 'js-codemod/transforms/object-shorthand'},
   {path: 'js-codemod/transforms/no-vars'},
+  {path: 'shopify-codemod/transforms/global-identifier-to-import'},
   {path: 'js-codemod/transforms/unquote-properties'},
 ];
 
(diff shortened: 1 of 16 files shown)


packages/esify/index.js
```diff
@@ -64,6 +67,21 @@ function loadOptions() {
...
+        moment: 'moment',
+        jstz: 'jstimezonedetect',
+        mousetrap: 'mousetrap',
+        URI: 'urijs',
+        URITemplate: 'urijs/src/URITemplate',
+        ReconnectingWebSocket: 'shopify-reconnecting-websocket',
+        d3: 'd3',
+        NProgress: 'NProgress',
+        FastClick: 'shopify-fastclick',
+        Clipboard: 'clipboard',
```
bouk: nice :+1:

bouk:
👌

packages/shopify-codemod/README.md
```diff
@@ -12,6 +12,29 @@ This repository contains a collection of Codemods written with [JSCodeshift](htt
 
 ## Included Transforms
 
+### `global-identifer-to-import`
+
+Transforms global identifiers that you specify into the appropriate import statements. In order for this to work, you must set the `globalIdentifiers` option to an object where the keys are the names of globals used in your script, and the keys are the import paths for those globals.
```
GoodForOneFare: `Transforms global identifiers` seems misleading because this is an additive operation.

```diff
-the keys are the import paths
+the values are the import path
```

Overall, it could be more concise.  ```Creates import statements for global references.   Use the `globalIdentifiers` option to specify identifier/import path pairs.```

GoodForOneFare:
Really nice stuff.  Once the key=>value typo is fixed, this is good to 🚢

packages/shopify-codemod/README.md
lemonmade: 👍

PR #128, opened 2016-05-24
Fix mocha context to closure on computed props

### LINKED ISSUE #94: Fix `mocha-context-to-closure` error when mixing setup/this/CallExpression
esify spits out this error:
```
undefined does not match field "name": string of type Identifier
  test/theme-editor-next-javascripts/functional/design_mode/shopify_api_next_test.coffee
```

The problem is in this `setup` code:
```
    for key, event of DesignMode.API.EVENTS
      @[key.toLowerCase()] = @spy()
      document.removeEventListener(event, @[key.toLowerCase()])
```

Things are fine if:
- `remove-useless-return-from-test` is disabled
- `@[key.toLowerCase()]` lines are changed to `foo[key.toLowerCase()]`
- The `toLowerCase()` call is removed

Minimal failing test input:
```
suite('DesignMode.ShopifyApi', function() {
  setup(function() {
    for (let [key, event] of foo) {
      this[key.toLowerCase()] = bar;
    }
  });
});
```

lemonmade:
Fixes #94. Wasn't checking for computed properties before :(

cc/ @GoodForOneFare @fandy

### DIFF
--- a/packages/shopify-codemod/test/fixtures/mocha-context-to-closure/dynamic-names.input.js
+++ b/packages/shopify-codemod/test/fixtures/mocha-context-to-closure/dynamic-names.input.js
@@ -1,0 +1,13 @@
+suite('foo', function() {
+  setup(function() {
+    this[bar] = () => 'baz';
+  });
+
+  teardown(function() {
+    this[bar.toLowerCase()] = 'baz';
+  });
+
+  it('works', function() {
+    expect(this[bar.toLowerCase()]).to.equal('baz');
+  });
+});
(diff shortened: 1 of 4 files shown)


GoodForOneFare:
LGTM 👍 

Sidebar: I can never remember what this transform does from the name alone 😕 If we can come up with something better, I'd be happy to PR a JS+docs update.  `mocha-this-to-suite-scope`?

lemonmade:
lol, I am pretty ambivalent about the name, if you feel strongly I have no problem with renaming it.

Issue #86, opened 2016-05-17
Convert default export objects to named exports


lemonmade:
```js
export default {
  foo: 'bar',
  baz: qux,
};
```

would be better written as:

```js
export const foo = 'bar';
export const baz = qux;
```

fandy:
Fixed in #129

PR #129, opened 2016-05-24
Convert default export objects to named exports

fandy:
Deconstructs default exports as requested in https://github.com/Shopify/javascript/issues/86

/cc @lemonmade @GoodForOneFare


packages/shopify-codemod/test/transforms/convert-default-export-objects-to-named-exports.test.js
```diff
@@ -0,0 +1,11 @@
+import 'test-helper';
+import convertDefaultExportObjectsToNamedExports from 'convert-default-export-objects-to-named-exports';
+
+describe('convertDefaultExportObjectsToNamedExports', () => {
+  it('converts export object literals', () => {
+    expect(convertDefaultExportObjectsToNamedExports).to.transform('convert-default-export-objects-to-named-exports/literals');
+  });
+  it('converts export object variables', () => {
```
lemonmade: I know this is so knitpicky, but I always prefer a space between calls like this :)

packages/shopify-codemod/transforms/convert-default-export-objects-to-named-exports.js
```diff
@@ -0,0 +1,21 @@
+export default function convertDefaultExportObjectsToNamedExports({source}, {jscodeshift: j}, {printOptions = {quote: 'single'}}) {
+
+  return j(source)
+    .find(j.ExportDefaultDeclaration)
+    .forEach((nodePath) => {
+      const exportIndex = nodePath.parentPath.node.body.findIndex((prop) => prop.type === 'ExportDefaultDeclaration');
+      const body = nodePath.parentPath.value;
+      delete body[exportIndex];
```
lemonmade: I think it's better to do `nodePath.replace()` here, and should have the same effect.

packages/shopify-codemod/transforms/convert-default-export-objects-to-named-exports.js
```diff
@@ -0,0 +1,21 @@
+export default function convertDefaultExportObjectsToNamedExports({source}, {jscodeshift: j}, {printOptions = {quote: 'single'}}) {
+
+  return j(source)
+    .find(j.ExportDefaultDeclaration)
+    .forEach((nodePath) => {
+      const exportIndex = nodePath.parentPath.node.body.findIndex((prop) => prop.type === 'ExportDefaultDeclaration');
```
lemonmade: It's probably more clear to do `nodePath.parentPath.node.body.indexOf(nodePath.node)`.

packages/shopify-codemod/transforms/convert-default-export-objects-to-named-exports.js
```diff
@@ -0,0 +1,21 @@
...
+      const body = nodePath.parentPath.value;
```
lemonmade: You can store this to use in the previous line as well.

packages/shopify-codemod/transforms/convert-default-export-objects-to-named-exports.js
```diff
@@ -0,0 +1,21 @@
...
+      delete body[exportIndex];
+      const declarations = nodePath.node.declaration.properties;
```
lemonmade: I prefer the `get` method for traversing: `nodePath.get('declaration', 'properties').node`

lemonmade: Also, will this die on a default export that does not have `properties`?

packages/shopify-codemod/transforms/remove-trailing-else-undefined-return.js
```diff
@@ -5,10 +5,10 @@ export default function removeTrailingElseUndefinedReturn({source}, {jscodeshift
   return j(source)
       .find(j.Function, {
         body: {
-          body: matchLast({
+          body: matchLast(j, {
             type: 'IfStatement',
             alternate: {
-              body: matchLast({
+              body: matchLast(j, {
```
lemonmade: Do we still need this?

packages/shopify-codemod/test/fixtures/convert-default-export-objects-to-named-exports/literals.input.js
```diff
@@ -0,0 +1,9 @@
+export default {
```
lemonmade: Three sanity tests I'd like to see:

- default exports that do not export objects (as noted below)
- named exports that export an object (unchanged)
- objects with computed properties (will have to stay unchanged since we don't know the name of the export statically).

packages/shopify-codemod/test/fixtures/convert-default-export-objects-to-named-exports/literals.input.js
```diff
@@ -0,0 +1,9 @@
...
+  foo: 'bar',
+  sho: [1, 2, 3],
+  cho: {
+  	foo: 'dee',
```
GoodForOneFare: No tabs, please.

packages/shopify-codemod/transforms/convert-default-export-objects-to-named-exports.js
```diff
@@ -0,0 +1,21 @@
...
+
+      let addedIndex = 0;
```
GoodForOneFare: You could just do `for (const dec of declarations.reverse())`.

packages/shopify-codemod/transforms/convert-default-export-objects-to-named-exports.js
```diff
@@ -0,0 +1,21 @@
+export default function convertDefaultExportObjectsToNamedExports({source}, {jscodeshift: j}, {printOptions = {quote: 'single'}}) {
+
+  return j(source)
+    .find(j.ExportDefaultDeclaration)
+    .forEach((nodePath) => {
```
GoodForOneFare: Destructuring saves a few chars later on: `.forEach(({parentPath, node}) => {`

packages/shopify-codemod/transforms/convert-default-export-objects-to-named-exports.js
```diff
@@ -0,0 +1,21 @@
...
+      const exportIndex = nodePath.parentPath.node.body.findIndex((prop) => prop.type === 'ExportDefaultDeclaration');
+      const body = nodePath.parentPath.value;
+      delete body[exportIndex];
+      const declarations = nodePath.node.declaration.properties;
+
+      let addedIndex = 0;
+      for (const dec of declarations) {
+        const varDec = j.variableDeclarator(dec.key, dec.value);
```
GoodForOneFare: The const per AST element was good for walking through the problem, but I think it's okay to compress it down to this now:
```js
        const namedExport = j.exportNamedDeclaration(
          j.variableDeclaration('const', [
            j.variableDeclarator(dec.key, dec.value)
          ])
        );
```

packages/shopify-codemod/transforms/convert-default-export-objects-to-named-exports.js
GoodForOneFare: Ah, nevermind.  Just saw Chris' other comments and this doesn't make sense.

packages/shopify-codemod/test/fixtures/convert-default-export-objects-to-named-exports/literals.input.js
lemonmade: I still don't think these tests. Looking for ones that check these cases:

```js
// not an object
export default [1, 2, 3];
export default 12;

// named exports
export const myObj = {foo: 'bar'};

// default exports with computed props
export default {
  [foo]: 'bar',
  'baz': qux,
};
```

packages/shopify-codemod/transforms/remove-empty-returns.js
```diff
@@ -1,7 +1,24 @@
...
 
 export default function removeEmptyReturns({source}, {jscodeshift: j}, {printOptions = {}}) {
-  return j(source)
+  const sourceAST = j(source);
+  sourceAST
+    .find(j.IfStatement, {
+      alternate: {
+        body: [{
+          type: 'ExpressionStatement',
+          expression: {
```
lemonmade: Would be nicer to use our `isUndefined` utility here.

packages/shopify-codemod/test/fixtures/convert-default-export-objects-to-named-exports/literals.input.js
lemonmade: There should also be a test (and a bail condition) for properties that are functions that refer to `this`.

lemonmade:
```js
export default {
  foo() { return this.bar; }
}
```

Also, would be nice to export any functions as function declarations rather than expressions.

PR #130, opened 2016-05-24
Scripts to create/rename transforms

GoodForOneFare:
These have made my life easier, so I thought I'd share.  If they're not useful, I'll keep them in my .gitignore :)

`bin/create-transform my-neato-transform` - creates a transform, test file, and `basic` fixture files
`bin/rename-transform my-neato-transform my-coolio-transform` - renames the transform, test, and fixture dir

/cc @lemonmade, @fandy

### DIFF
--- a/packages/shopify-codemod/bin/create-transform
+++ b/packages/shopify-codemod/bin/create-transform
@@ -1,0 +1,7 @@
+#!/usr/bin/env node --harmony_destructuring
+
+/* eslint-disable node/shebang */
+const create = require('../transform-cli/create-transform');
+
+const [,, dasherizedName] = process.argv;
+create(dasherizedName);
--- a/packages/shopify-codemod/transform-cli/create-transform.js
+++ b/packages/shopify-codemod/transform-cli/create-transform.js
@@ -1,0 +1,41 @@
+/* eslint-disable no-sync */
+const fs = require('fs');
+const path = require('path');
+const utils = require('./transform-management-utils');
+
+function createFileSync(filePath, content) {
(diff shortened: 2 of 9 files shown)


packages/shopify-codemod/transform-cli/create-transform.js
```diff
@@ -0,0 +1,42 @@
+/* eslint-disable no-sync */
+const fs = require('fs');
+const path = require('path');
+const utils = require('./transform-management-utils');
+
+function createFileSync(filePath, content) {
+  const fd = fs.openSync(filePath, 'wx');
```
lemonmade: Why do you need to open it before writing? I always just write, am I a bad person?

packages/shopify-codemod/bin/create-transform
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env node --harmony_destructuring
```
lemonmade: Want to `npmignore` the new so we don't publish them?

lemonmade:
Minor comments, thanks for the utilities :) additions to the README might be nice too, unless I just missed them.

packages/shopify-codemod/transform-cli/create-transform.js
GoodForOneFare: It's weird because I write directly to file in the rename helper.  Who knows what I was thinking?  Fixed now.

GoodForOneFare:
NPM ignored, file writes fixed, README updated.  Thanks, @lemonmade!

PR #131, opened 2016-05-24
Improve global to import transform

### LINKED ISSUE #95: Make `global-reference-to-import` faster
Based on a facile profile of transforms, `global-reference-to-import` accounts for ~60% of `esify`'s runtime.  

Most likely, this is caused by continually searching the codebase with `ag`/`ack`.  Adding a cache may significantly speed this up.

-----

Timing based on a full transform of `shopify/shopify/**/*.coffee`, with `ag`.

| Transform | Time (minutes) |
|----------------|----------------------|
| shopify-codemod/global-reference-to-import | 11.65 |
| js-codemod/no-vars | 1.54 |
| shopify-codemod/mocha-context-to-closure | 0.35 |
| shopify-codemod/global-identifier-to-import | 0.34 |
| shopify-codemod/coffeescript-soak-to-condition | 0.32 |
| shopify-codemod/function-to-arrow | 0.32 |
| shopify-codemod/remove-useless-return-from-test | 0.31 |
| shopify-codemod/ternary-statement-to-if-statement | 0.31 |
| js-codemod/unquote-properties | 0.27 |
| shopify-codemod/coffeescript-range-output-to-helper | 0.27 |
| shopify-codemod/conditional-assign-to-if-statement | 0.27 |
| shopify-codemod/remove-addeventlistener-returns | 0.27 |
| shopify-codemod/constant-function-expression-to-statement | 0.26 |
| js-codemod/template-literals | 0.25 |
| shopify-codemod/strip-template-literal-parenthesis | 0.25 |
| shopify-codemod/remove-trailing-else-undefined-return | 0.25 |
| js-codemod/object-shorthand | 0.25 |
| shopify-codemod/remove-empty-returns | 0.24 |
| shopify-codemod/mocha-context-to-global-reference | 0.17 |
| shopify-codemod/global-assignment-to-default-export | 0.16 |
| js-codemod/arrow-function | 0.14 |
| shopify-codemod/assert/call-count-equals-to-assert-called | 0.14 |
| shopify-codemod/assert/called-method-to-assert-called | 0.13 |
| shopify-codemod/assert/move-literals-to-expected-argument | 0.12 |
| shopify-codemod/assert/called-with-methods-to-assert-called-with | 0.12 |
| shopify-codemod/assert/negated-assert-ok-to-assert-not-ok | 0.12 |
| shopify-codemod/assert/assert-to-assert-ok | 0.12 |
| shopify-codemod/assert/assert-false-to-assert-fail | 0.12 |
| shopify-codemod/assert/called-equals-boolean-to-assert-called | 0.12 |
| shopify-codemod/assert/equality-comparisons-to-assertions | 0.12 |
| shopify-codemod/assert/falsy-called-method-to-assert-not-called | 0.11 |

### LINKED ISSUE #99: A file shouldn't try to import itself
`admin/lib/money.js` tries to ```import Money from `admin/lib/money````:
```
'expose Shopify.Money';
import _ from 'lodash';
import Number from 'admin/lib/number';
import CurrencyDB from 'admin/currency_db.coffee';
import Money from 'admin/lib/money';
const MONEY_FORMAT_REGEX = /(\{\{\s*(\w+)\s*\}\})/;
const COMMON_DELIMITERS = '\',.';

export default class Money {
```

Also affected: `admin/lib/time`, `admin/modules/domains/domain-form`.

lemonmade:
Fixes #99. Maybe fixes #95. This PR fixes files importing themselves by checking the `expose` directive (so, it must come after the default export transform, which it does), and adds a global cache that I believe will speed up the lookups between transforms.

### DIFF
--- a/packages/shopify-codemod/transforms/global-reference-to-import.js
+++ b/packages/shopify-codemod/transforms/global-reference-to-import.js
@@ -14,6 +14,8 @@ function determineFileSearcher() {
   }
 }
 
+const fileForIdentifier = {};
+
 export default function globalReferenceToImport(
   {source},
   {jscodeshift: j},
@@ -23,7 +25,11 @@ export default function globalReferenceToImport(
     appGlobalIdentifiers,
   }) {
   const [binary, args] = determineFileSearcher();
-  const fileForIdentifier = {};
+
+  const sourceLocation = resolve(javascriptSourceLocation);
+  function relativePath(file) {
+    return relative(sourceLocation, file).replace(/\.[a-z]+$/, '');
+  }
 
   /*
   * findDeclaringFile uses ack or the_silver_searcher to find the file in which something is declared.
@@ -51,7 +57,7 @@ export default function globalReferenceToImport(
       throw new Error(`Found multiple definitions for ${identifier}`);
     }
 
-    return relative(absolutePath, files[0]).replace(/\.[a-z]+$/, '');
+    return files[0];
   }
 
   function getDeclaringFile(identifier) {
@@ -70,10 +76,25 @@ export default function globalReferenceToImport(
   return j(source)
     .find(j.Program)
     .forEach((path) => {
+      const exposeDirective = path.get('body').filter(({node}) => j.match(node, {
+        expression: {
+          type: j.Literal.name,
+          value: (value) => typeof value === 'string' && value.startsWith('expose'),
+        },
+      }))[0];
+
+      const exposed = exposeDirective && exposeDirective.get('expression').node.value.replace(/^expose\s+/, '');
+      const exposedLocalName = exposed && findLastMember(j(exposed).paths()[0].get('program', 'body', 0, 'expression').node).name;
+
       const imports = new Map();
 
       function registerMember(node) {
         const member = j(node).toSource();
+
+        if (member === exposed) {
+          return exposedLocalName;
+        }
+
         if (imports.has(member)) {
           return imports.get(member).name;
         } else {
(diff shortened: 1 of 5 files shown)


bouk:
nice :+1:

packages/shopify-codemod/transforms/global-reference-to-import.js
```diff
@@ -131,7 +152,7 @@ export default function globalReferenceToImport(
           body,
           j.importDeclaration([
             j.importDefaultSpecifier(j.identifier(name)),
-          ], j.literal(file))
+          ], j.literal(relativePath(file)))
```
GoodForOneFare: I think the snake case name will make it through to imports here.

GoodForOneFare: My bad, the snake case is getting through via `imports.set` at line 104.

```
          const file = getDeclaringFile(member);
          if (file === null) { return null; }
          const name = findLastMember(node).name;
          imports.set(member, {file, name});  <------------ here.
          return name;
```

GoodForOneFare:
Code looks great.  Just the snake case thing to fix before 🚢 

I did a quick esify runthrough and everything was done in *less time than the global import transform used to take on its own*!! That's an outstanding result for a few shuffled lines 😁  Thank you!

packages/shopify-codemod/transforms/global-reference-to-import.js
GoodForOneFare: Okay, I was reading the diff the wrong way round 😥 This is really fixing some imports where the snake case version was sneaking in.  That's good!

Before/after this transform:
```diff
-import ApplePayCertificatePoller from 'admin/modules/apple_pay_certificate_poller';
+import ApplePayCertificatePoller from 'admin/modules/apple-pay-certificate-poller';
```

Issue #109, opened 2016-05-23
Remove empty catch blocks


GoodForOneFare:
Remove the empty catch block from code like this:
```js
// Before
try {
  foo();
} catch (undefined) {} finally {
  bar();
}

// After
try {
  foo();
} finally {
  bar();
}

// Before
try {
  foo();
} catch (e) {} finally {
  bar();
}

// After
try {
  foo();
} finally {
  bar();
}
```

fandy:
@lemonmade and I decided it's easier to do this in another code-mod transform instead of Decaf (for the sake of sanity).

fandy:
Fixed in #153

PR #132, opened 2016-05-24
`catch (undefined)` => `catch (error)`

GoodForOneFare:
Decaf generates many of these :/  Note there's a task (#109) to remove empty `catch` blocks.  But that's a PR for another day :)

/cc @lemonmade, @fandy

### DIFF
--- a/packages/shopify-codemod/transforms/catch-undefined-to-catch-error.js
+++ b/packages/shopify-codemod/transforms/catch-undefined-to-catch-error.js
@@ -1,0 +1,17 @@
+export default function catchUndefinedToCatchError({source}, {jscodeshift: j}, {printOptions = {quote: 'single'}}) {
+  function hasNoErrorReferences(path) {
+    const usage = j(path).find(j.Identifier, {name: 'error'});
+
+    return usage.size() === 0;
+  }
+
+  return j(source)
+    .find(j.CatchClause, {
+      param: {
+        name: 'undefined',
+      },
+    })
+    .filter(hasNoErrorReferences)
+    .forEach(({node: {param}}) => { param.name = 'error'; })
+    .toSource(printOptions);
+}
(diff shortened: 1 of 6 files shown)


packages/shopify-codemod/transforms/catch-undefined-to-catch-error.js
```diff
@@ -0,0 +1,17 @@
+export default function catchUndefinedToCatchError({source}, {jscodeshift: j}, {printOptions = {quote: 'single'}}) {
+  function hasNoErrorReferences(path) {
+    const usage = j(path).find(j.Identifier, {name: 'error'});
```
GoodForOneFare: I could have done something more involved with scope/binding info here, but ultimately `error` shadowing isn't an issue in `shopify/shopify`, and the linter will flag it if it becomes an issue.

packages/shopify-codemod/transforms/catch-undefined-to-catch-error.js
```diff
@@ -0,0 +1,17 @@
...
+    return usage.size() === 0;
+  }
+
+  return j(source)
+    .find(j.CatchClause, {
+      param: {
+        name: 'undefined',
+      },
+    })
+    .filter(hasNoErrorReferences)
```
fandy: Is it better to keep `hasNoErrorReferences` inline since it's only used once?

lemonmade:
Code LGTM. Is this something that should be fixed in decaf, or is it appropriate to have this as the default catch clause?

GoodForOneFare:
Damn, that's a good question.  Digging into the code, defaulting to `error` and omitting the catch entirely if a finally exists should be easy.

Thanks for pointing me in the right direction :)

packages/shopify-codemod/transforms/catch-undefined-to-catch-error.js
GoodForOneFare: We're not consistent about this, but my preference is to reduce cognitive load by splitting more elaborate logic into functions with human readable names.  That way, someone can read through the main flow of the transform without having to unpack every single detail.

I didn't add `hasNoErrorReferences` to `utils` because:
- It deals with an edge case generated by `decaf`, and is unlikely to function correctly elsewhere
- It's best to let things propagate up to `utils` once we see it used in multiple locations; if something is never reused, it's more readable to keep it close to its caller

Issue #104, opened 2016-05-23
Clean up `no-return-assign` errors


GoodForOneFare:
Return-assignments should be split into two lines.
```js
// Before
Bindings.afterBound(() => {
  return this.lastInput = this.input;
});

// After
Bindings.afterBound(() => {
  this.lastInput = this.input;
  return this.lastInput;
});
```

```js
// Before
setInput(newInput) {
  return this.input = this.lastInput = newInput;
}

// After
setInput(newInput) {
  this.input = this.lastInput = newInput;
  return this.input;
}
```

```js
// Before
return templates[templateName] = el.innerHTML;

// After
templates[templateName] = el.innerHTML;
return templates[templateName];
```

fandy:
Fixed in https://github.com/Shopify/javascript/pull/134 `splitReturnAssignments`

PR #134, opened 2016-05-25
Split return assignments

fandy:
This transform addresses https://github.com/Shopify/javascript/issues/104

/cc @lemonmade @GoodForOneFare


packages/shopify-codemod/test/transforms/split-return-assignments.js
```diff
@@ -0,0 +1,12 @@
+import 'test-helper';
```
GoodForOneFare: These tests aren't being executed.  Needs the `.test.js` suffix.

packages/shopify-codemod/transforms/splitReturnAssignments.test.js
```diff
@@ -0,0 +1,37 @@
...
+          argument: {
+            type: 'AssignmentExpression',
+          },
+        }),
+      },
+    })
+    .forEach(({node: {body: {body}}}) => {
+
+      const {left, right} = body[body.length - 1].argument;
+      delete body[body.length - 1];
```
lemonmade: `body.pop()`

lemonmade: I think that also returns the popped value, from which you could grab the `{left, right}`, if you want.

packages/shopify-codemod/test/transforms/split-return-assignments.js
fandy: Ah I added it to the wrong file. Thanks for catching. 👍

packages/shopify-codemod/transforms/splitReturnAssignments.test.js
```diff
@@ -0,0 +1,37 @@
...
+
+      body.push({
```
lemonmade: Pretty sure you should be using the builders (`j.expressionStatement()`, etc) here, instead of object literals.

packages/shopify-codemod/test/fixtures/split-return-assignments/basic.output.js
```diff
@@ -0,0 +1,5 @@
+foo(() => {
+  console.log(bar);
+  this.sho = this.doo;
+  return this.sho;
+});
```
lemonmade: You'll probably also want a test for

```js
foo(() => {
  return foo = bar = baz;
});
```

and:

```js
foo(() => foo = bar);
```

packages/shopify-codemod/transforms/splitReturnAssignments.test.js
fandy: Assigning the return value of `body.pop` to `{left, right}` is very clever! Updated.

GoodForOneFare:
I don't understand the change to `remove-useless-return-from-test.js`.

GoodForOneFare:
I think there's something wrong with your git workflow.  Are you creating branches from your previous PR's branch?  It's not a huge deal in a project of this size, but you may get an instant 👎 on a commit history like this in a `shopify/shopfiy` PR.

Let me or @lemonmade know when you're starting a new PR, and we can work this out together :)

GoodForOneFare:
@fandy would you consider adding a pre-push hook to prevent failing code from coming up for review, please?

In your `shopify/javascript` checkout dir:
```
echo 'cd packages/shopify-codemod && npm test' > .git/hooks/pre-push
```

This'll run the codemod tests, every time you try to `git push`, and back out if the tests fail.

packages/shopify-codemod/test/fixtures/split-return-assignments/basic.output.js
fandy: Do I have to account for 

```
foo(() => {
  return foo = bar = baz = faz = taz = 0;
});
```

lemonmade: Yeah, any group of assignments should be handled (they will all be handled the same way anyways).

fandy:
Added a few more test cases. @lemonmade how does it look?

packages/shopify-codemod/transforms/split-return-assignments.js
```diff
@@ -0,0 +1,112 @@
...
+          argument: {
+            type: 'AssignmentExpression',
+          },
+        }),
+      },
+    })
+    .forEach(({node: {body: {body}}}) => {
+      const {left, right} = body[body.length - 1].argument;
+      const returnLine = body[body.length - 1];
+      if (returnLine.argument.left) {
```
lemonmade: When would an assignment expression not have a `left`?

packages/shopify-codemod/transforms/split-return-assignments.js
```diff
@@ -0,0 +1,112 @@
...
+        delete body[body.length - 1];
+        body.push({
```
lemonmade: Use the builder functions (`j.expressionStatement`) instead of object literals.

packages/shopify-codemod/transforms/split-return-assignments.js
```diff
@@ -0,0 +1,112 @@
...
+          type: 'ExpressionStatement',
+          expression: {
+            type: 'AssignmentExpression',
+            operator: '=',
```
lemonmade: This will break if the original operator was not a `=` (for example, `return foo += bar`). Can you make this use the original expression's operator, and add a test for non-`=` operators?

packages/shopify-codemod/transforms/split-return-assignments.js
```diff
@@ -0,0 +1,112 @@
...
+            left,
+            right,
+          },
+        });
+
+        if (returnLine.argument.left.name) {
+          body.push({
```
lemonmade: You don't need all those cases, you should be able to just have a single case, where you do `body.push(j.returnStatement(returnLine.argument.left))`, which will work regardless of the type of expression that was in the `return` statement.

packages/shopify-codemod/transforms/split-return-assignments.js
```diff
@@ -0,0 +1,112 @@
...
+            },
+          },
+        });
+      }
+    });
+  sourceAST
+    .find(j.ExpressionStatement, {
+      type: 'ExpressionStatement',
+      expression: {
+        type: 'CallExpression',
```
lemonmade: I don't think you want to go to the expression level here, you just want to find a `j.ArrowFunctionExpression` with the body being just the assignment expression.

packages/shopify-codemod/transforms/split-return-assignments.js
```diff
@@ -0,0 +1,112 @@
...
+            body: {
+              type: 'AssignmentExpression',
+              operator: '=',
+            },
+          },
+        ],
+      },
+    }).forEach(({node: {expression}}) => {
+      const {left, right} = expression.arguments[0].body;
+      expression.arguments[0].body = {
```
lemonmade: Same comments about using the builders rather than literals.

packages/shopify-codemod/transforms/split-return-assignments.js
```diff
@@ -0,0 +1,112 @@
...
+          type: 'ExpressionStatement',
+          expression: {
+            type: 'AssignmentExpression',
+            operator: '=',
+            left,
+            right,
+          },
+        });
+
+        if (returnLine.argument.left.name) {
```
lemonmade: If you find yourself reaching deep into a nested property multiple times, you probably want to pull that property up as its own variable in scope/

fandy:
Updated.

packages/shopify-codemod/test/fixtures/split-return-assignments/multi.output.js
```diff
@@ -0,0 +1,14 @@
...
+  this.sho = this.doo;
+  this.doo = this.zoo;
+  this.zoo = 0;
+  return this.sho;
+});
+
+foo(() => {
+  foo();
+  foo = bar;
+  bar = baz;
```
lemonmade: ⚠️ this is not equivalent to `return foo = bar = baz;`. In that case, `foo` will equal whatever `baz` is. In this case, it will equal whatever `bar` is. We should just move the assignment up altogether rather than splitting it up.

fandy: Oops - now fixed with the correct order. :)

lemonmade: I still lean to just moving the whole assignment up rather than splitting it into individual assignments. I'm not 100% sold on that, though — what do you think?

fandy:
@lemonmade is this PR ready to merge?

packages/shopify-codemod/test/fixtures/split-return-assignments/multi.output.js
lemonmade: Order is still wrong on these.

packages/shopify-codemod/test/fixtures/split-return-assignments/this.output.js
```diff
@@ -0,0 +1,5 @@
+foo(() => {
+  console.log(bar);
+  this.sho = this.doo = this.zoo;
```
lemonmade: Why is this one moving the full assignment up, where the other ones that have more than two assignments break it down?

lemonmade: This test isn't actually being run, so presumably that's why :P

packages/shopify-codemod/transforms/split-return-assignments.js
lemonmade: Still not resolved, so this will still not work for something like `const foo = () => bar = baz;`

packages/shopify-codemod/transforms/split-return-assignments.js
```diff
@@ -0,0 +1,52 @@
...
+        }),
+      },
+    })
+    .forEach(({node: {body: {body}}}) => {
+      const returnLine = body[body.length - 1];
+      forEachAssignment(body[body.length - 1].argument, (assignment) => {
+        body.splice(body.indexOf(returnLine) + 1, 0, j.expressionStatement(assignment));
+      });
+      delete body[body.indexOf(returnLine)];
+      body.push(j.returnStatement(returnLine.argument.left));
```
lemonmade: I'd still really prefer sticking to `NodePath`s instead if dropping down into the nodes, seems like it should be doable here.

packages/shopify-codemod/transforms/split-return-assignments.js
```diff
@@ -0,0 +1,52 @@
...
+      body: {
+        body: matchLast(j.ReturnStatement, {
+          type: 'ReturnStatement',
+          argument: j.AssignmentExpression,
+        }),
+      },
+    })
+    .forEach(({node: {body: {body}}}) => {
+      const returnLine = body[body.length - 1];
+      forEachAssignment(body[body.length - 1].argument, (assignment) => {
```
lemonmade: You still haven't addressed the comment I made about disliking the "callback-for-each-item" style being used here.

Issue #100, opened 2016-05-23
Fix conditional assignments


GoodForOneFare:
The `operation.done` callback contains this CoffeeScript assignment:
```coffeescript
      if @nextPageParams?
        @title = "Load next #{@LARGE_PAGE_SIZE} products"

```

The transformed JS is missing a `this.title` assignment:
```js
      if (this.nextPageParams != null) {
        this.nextPageParams;
      }
```

GoodForOneFare:
A couple of other affected files:

`bulk_operations.coffee`
```coffeescript
data['value'] = operationData if operationData?
```

```js
if (operationData != null) {
  operationData;
}
```

`two_factor_auth.coffee`:
```coffeescript
password = $(form).find('input[name=password]').val()
@password = password if password?
```

```js
var password = $(form).find('input[name=password]').val();

if (password != null) {
  password;
}
```

PR #135, opened 2016-05-25
Fix existential soak checks

### LINKED ISSUE #97: Fix soaked conditional assigments
`admin_embedded_app_host.coffee` contains a couple of conditional assignments like:
```
pagination.previous = Shopify.EmbeddedAppButtons.build(pagination.previous) if pagination?.previous?
```

This gets rewritten to a complex `if` that doesn't assign to `pagination.previous`:

```
    if (((typeof pagination !== 'undefined' && pagination !== null ? pagination.previous : void 0)) != null) {
      if (typeof pagination !== 'undefined' && pagination !== null) {
        pagination.previous;
      } else {
        void 0;
      }
    }
```

lemonmade:
Fixes #97. This PR updates `coffeescript-soak-to-condition` to understand an additional edge case with soak/ existential operators.

cc/ @GoodForOneFare @fandy


lemonmade:
Also fixes #100.

GoodForOneFare:
👍

PR #136, opened 2016-05-25
Remove block statements for single-return arrow functions

lemonmade:
This PR augments `function-to-arrow` to also remove block statements from existing single-return arrow functions, which are generated by other things and aren't currently being handled.

### DIFF
--- a/packages/shopify-codemod/transforms/function-to-arrow.js
+++ b/packages/shopify-codemod/transforms/function-to-arrow.js
@@ -11,7 +11,19 @@ export default function functionToArrow({source}, {jscodeshift: j}, {printOption
     return !isMember(path) && !containsThisExpression(path);
   }
 
-  return j(source)
+  const root = j(source);
+
+  root
+    .find(j.ArrowFunctionExpression, {
+      body: {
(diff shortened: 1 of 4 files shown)


GoodForOneFare:
Should one line return-assignments stay inside a block?

packages/shopify-codemod/transforms/function-to-arrow.js
```diff
@@ -11,7 +11,19 @@ export default function functionToArrow({source}, {jscodeshift: j}, {printOption
     return !isMember(path) && !containsThisExpression(path);
   }
 
-  return j(source)
+  const root = j(source);
+
+  root
```
fandy: Should we use `const sourceAST` as a convention for caching the AST, since other files are currently using `sourceAST`?

lemonmade: 👍

lemonmade:
@GoodForOneFare @fandy I think the codemod we have to handle return assignments should just handle block-less functions; it removes any order dependencies, and makes it more resilient to potentially unexpected decaf/ other transformations.

Issue #126, opened 2016-05-24
Files that mix `$` and `jQuery` try to `import 'jquery'` twice


GoodForOneFare:
Files like `skip-link.coffee` reference `jQuery`/`$`, which generates duplicate `import` statements.
```
$(document).on('keydown', '.js-btn-skip', evt => {
  if (evt.keyCode === jQuery.ui.keyCode.ENTER) {
```

`combobox` is a special case that ends up with 3 jQuery definitions.

```js
import jQuery from 'jquery';
import $ from 'jquery';
// ...SNIP...
const $ = jQuery;
```

lemonmade:
What do you think the correct solution is here? I'm thinking maybe an earlier transform to rename `jQuery` to `$` and a special fix for `combobox`, but open to suggestions.

GoodForOneFare:
Yes, exactly that.  There's way more instances of `$` in core, and there shouldn't be any difference in behaviour between `$`/`jQuery`.

PR #139, opened 2016-05-26
Add rename-identifier and rename-property

### LINKED ISSUE #102: Use es-like lodash names
Lodash's docs show that the following `_` methods can be safely renamed to their non-aliased versions.

| Old name | Standard name |
|-----------------|---------------------|
| _.first | _.head |
| _.each | _.forEach |
| _.eachRight | _.forEachRight |
| _.entries | _.toPairs |
| _.entriesIn | _.toPairsIn |
| _.extend | _.assignIn |
| _.extendWith | _.assignInWith |

lemonmade:
Fixes #102 and #126. This PR adds two new transforms, `rename-identifier` and `rename-property`, to handle arbitrary renaming of identifiers and member expressions, respectively.

cc/ @GoodForOneFare @fandy

### DIFF
--- a/packages/shopify-codemod/transforms/rename-property.js
+++ b/packages/shopify-codemod/transforms/rename-property.js
@@ -1,0 +1,18 @@
+import {pathIsFirstMember} from './utils';
+
+export default function renameProperty({source}, {jscodeshift: j}, {printOptions = {quote: 'single'}, renameProperties = {}}) {
+  function hasPropertyThatShouldBeRenamed({node: {computed, object, property}}) {
+    return !computed && renameProperties[object.name].hasOwnProperty(property.name);
+  }
+
+  if (Object.keys(renameProperties).length === 0) { return null; }
+
+  return j(source)
+    .find(j.MemberExpression, {object: {name: (name) => renameProperties.hasOwnProperty(name)}})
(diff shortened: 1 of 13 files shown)


packages/shopify-codemod/transforms/rename-property.js
```diff
@@ -0,0 +1,18 @@
+import {pathIsFirstMember} from './utils';
+
+export default function renameProperty({source}, {jscodeshift: j}, {printOptions = {quote: 'single'}, renameProperties = {}}) {
+  function hasPropertyThatShouldBeRenamed({node: {computed, object, property}}) {
+    return !computed && renameProperties[object.name].hasOwnProperty(property.name);
```
GoodForOneFare: Could you add a sanity test for the `!computed`, please?

lemonmade: I think `_[each]()` tests this, does it not?

GoodForOneFare:
Nice and concise!  Looks great 👍 🚢 

It looks like we don't have to worry about translating `window.$`, etc because any files doing that have/should be converted to `.js` manually.

packages/shopify-codemod/transforms/rename-property.js
GoodForOneFare: It sure does!  Ignore me :)

PR #141, opened 2016-05-26
Add a computed-literal-keys-to-dot-notation transform

### LINKED ISSUE #111: Clean up `dot-notation` warnings
This is probably way tricker than it looks, or there may already be a mod for this in the wild.  Either way, search and discuss before starting work on this.

```js
// Before
if (this.foo['bar']) {
   qux();
}

const buzz = this.foo['foobar'];

// After 
if (this.foo.bar) {
   qux();
}

const buzz = this.foo.foobar;
```

lemonmade:
Fixes #111. This PR cleans up string literal keys that are valid identifiers to use dot notation for member expressions. Also fixed a potential bug in rename-property for renaming a property to a name that is not a valid identifier (unlikely, but whatever, may as well check).

cc/ @GoodForOneFare @fandy

### DIFF
--- a/packages/shopify-codemod/transforms/utils.js
+++ b/packages/shopify-codemod/transforms/utils.js
@@ -44,6 +44,12 @@ export function pathIsFirstMember({node, parentPath: {node: parentNode}}) {
   return !j.MemberExpression.check(parentNode) || parentNode.object === node;
 }
 
+const IDENTIFIER_REGEX = /^[a-zA-Z_$][a-zA-Z0-9_$]*$/;
+
+export function isValidIdentifier(identifier) {
+  return typeof identifier === 'string' && IDENTIFIER_REGEX.test(identifier);
+}
+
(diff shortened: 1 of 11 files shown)


packages/shopify-codemod/transforms/utils.js
```diff
@@ -44,6 +44,12 @@ export function pathIsFirstMember({node, parentPath: {node: parentNode}}) {
   return !j.MemberExpression.check(parentNode) || parentNode.object === node;
 }
 
+const IDENTIFIER_REGEX = /^[a-zA-Z_$][a-zA-Z0-9_$]*$/;
+
+export function isValidIdentifier(identifier) {
+  return typeof identifier === 'string' && IDENTIFIER_REGEX.test(identifier);
```
GoodForOneFare: Interesting contrast to the hoops that `unquote-properties` jumps through :)

GoodForOneFare:
Really good stuff 👍 Keep 'em coming.

PR #143, opened 2016-05-27
Add Node Babel presets

### LINKED ISSUE #39: Add a shopify-node Babel preset and ESLint config
For node projects, it would be nice to have babel transpiling or ESLint complaining about features that are not available in the version of Node/ npm we are trying to use wherever possible.

lemonmade:
This PR fixes #39. It's basically a problem only I really have; I write a bunch of little Node tools, and want to use our recommended Babel preset/ language features. Our default Babel preset needlessly transpires a lot of features for older versions of Node. This PR adds a bunch of version-specific Node Babel presets, accessible by extending the `shopify/node/<version>` preset in your Babel config. It also moves the React preset out of our default one, and instead provides a React-specific preset at `shopify/react`.

cc/ @bouk

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,14 @@
 # Changelog
 
 ## [Unreleased]
+### Added
+- Added node.js-specific Babel presets to `babel-preset-shopify`.
 
 ### Fixed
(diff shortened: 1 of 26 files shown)


lemonmade:
Note to self: think about whether this: https://github.com/rtsao/babel-preset-es2015-node is a better approach.

PR #144, opened 2016-05-27
eslint spaced-comment: Allow sprockets directives

bouk:
@lemonmade


bouk:
eslint complains if you use a sprockets directive, like

```
//= require_tree .
```

This make it recognize `=` as a special character that comes before the directive

lemonmade:
Thanks :+1:

PR #145, opened 2016-05-30
Add missing radix transform

### LINKED ISSUE #117: Add `10` radix to `parseInt` calls
Per the [MDN parseInt docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt):
> Always specify this parameter to eliminate reader confusion and to guarantee predictable behavior. Different implementations produce different results when a radix is not specified, usually defaulting the value to 10.

lemonmade:
Fixes #117. Does what it says on the tin.

cc/ @fandy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## [Unreleased]
+### Added
+- Added a `add-missing-parseint-radix` transform to add missing radix parameters to `parseInt` calls.
 
 
 ## [12.0.0]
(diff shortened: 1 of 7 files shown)


fandy:
Overall looks perfect, but if I'm being picky then this might be a valid test case:

```js
// before
parseInt('442', 10);
//after
parseInt(442, 10);
``

This way `parseInt` doesn't have to automatically do type coercion.

lemonmade:
@fandy I see where you're coming from, but I don't know that it's the best bet since 1) we have no instances of actually calling `parseInt` on a literal, only on references, and 2) we are basically just doing the same `parseInt` call ourselves to figure out what the numeric literal should be, which defeats the purpose of having the `parseInt` in the transformed code.

Issue #107, opened 2016-05-23
Clean up `shopify/jquery-dollar-sign-reference` warnings


GoodForOneFare:
Rename any locally declared vars that violate this rule.

```js
// Before
function foo() {
    const li = $('li.app-dropdown');
    li.removeClass('active');
    li.find('a').removeClass('active');
}

// After
function foo() {
    const $li = $('li.app-dropdown');
    $li.removeClass('active');
    $li.find('a').removeClass('active');
}
```

fandy:
The before and after examples here are the same.

GoodForOneFare:
Before has `li`, after has `$li`.

fandy:
This was more complicated than expected, so was put on hold in https://github.com/Shopify/javascript/pull/147

PR #147, opened 2016-05-30
Adds dollar sign to jQuery element declaration/ references

fandy:
Fixes https://github.com/Shopify/javascript/issues/107

/cc @lemonmade


lemonmade:
I don't think this really solves that particular issue. Even the example shows converted code that is incorrect (`console.log(bar)` not converted. This is also really easy to trick if you have common variable names in different scopes:

```js
function foo() {
  const bar = $('div');
  return bar;
}

function foo2() {
  const bar = 'something';
  return bar.toLowerCase();
}
```

becomes:

```js
function foo() {
  const $bar = $('div');
  return bar;
}

function foo2() {
  const bar = 'something';
  return $bar.toLowerCase();
}
```

I need to think about whether this transform is actually worth all the work it will take to do correctly, especially given the fact that it does not address many other violations of the linting rule (more complex jQuery expressions, jQuery nodes as members, etc).

fandy:
You're right, this is harder than it seems. I have not worked with jQuery in a while so maybe it's best for someone else to take over if this is still worth doing (or provide me with some more advanced test cases).

fandy:
Put on hold https://github.com/Shopify/javascript/issues/107#issuecomment-223292582

PR #149, opened 2016-05-31
Add implicit-coercion-to-explicit transform

### LINKED ISSUE #105: Clean up `no-implicit-coercion` warnings
```js
// Before
const hadErrorMessage = !!this.errorMessage;

// After
const hadErrorMessage = Boolean(this.errorMessage);
```

lemonmade:
Fixes #105.

cc/ @fandy


packages/shopify-codemod/test/transforms/implicit-coercion-to-explicit.test.js
```diff
@@ -0,0 +1,8 @@
+import 'test-helper';
+import transform from 'implicit-coercion-to-explicit';
```
fandy: Is it better to give the imported transform a more explicit name? Most of the other transforms currently use a camel cased version of the transform file.

fandy:
Overall looks good and ready. This PR reminded me to add transforms to Esify and the code-mod README.

packages/shopify-codemod/test/transforms/implicit-coercion-to-explicit.test.js
lemonmade: 👍

PR #152, opened 2016-06-01
Clean up esify and add eslint --fix

lemonmade:
This PR splits esify up into some more manageable chunks, and adds a final step of running `eslint --fix` on the transformed file. There are a pretty decent number of fixable linting violations that this should get rid of (most obviously: newline at the end of the file. Hurray!)

cc/ @bouk @fandy @justinthec


packages/esify/lib/eslint.js
```diff
@@ -0,0 +1,16 @@
...
+  return new Promise(function(resolve) {
+    var eslint = findESLintBinary();
+    if (eslint == null) { return; }
+    exec(eslint + ' ' + details.file + ' --fix', resolve);
+  });
+};
+
+function findESLintBinary() {
+  var possibleBinary = path.join(process.cwd(), 'node_modules/.bin/eslint');
+  return fs.statSync(possibleBinary).isFile() ? possibleBinary : null;
```
bouk: you can just do `npm bin` to find the correct directory

packages/esify/lib/codemod.js
```diff
@@ -1,6 +1,5 @@
 var path = require('path');
```
fandy: The `path` and `exec` module isn't used and is causing linting errors, causing tests to fail.

packages/esify/lib/codemod.js
```diff
@@ -1,6 +1,5 @@
...
-var decaf = require('shopify-decaf');
-var chalk = require('chalk');
+var exec = require('child_process').exec;
```
fandy: Delete this line as well.

packages/esify/lib/eslint.js
```diff
@@ -0,0 +1,16 @@
+var exec = require('child_process').exec;
+var path = require('path');
+var fs = require('fs');
+
+module.exports = function runESLint(details) {
+  return new Promise(function(resolve) {
```
fandy: I think we need the `es2015` preset on `babel-preset-shopify` in order to define `Promise`.

lemonmade: It was actually a linting environment thing that needed to be fixed. Thanks for catching that.

PR #153, opened 2016-06-02
Removes empty catch blocks

fandy:
Fixes https://github.com/Shopify/javascript/issues/109

/cc @lemonmade


packages/shopify-codemod/transforms/remove-empty-catch-blocks.js
```diff
@@ -0,0 +1,16 @@
...
+
+  return j(source)
+    .find(j.TryStatement, {
+      type: 'TryStatement',
+      handler: {
+        type: 'CatchClause',
+      },
+    })
+      .forEach(({node}) => {
+        if (node.handler.body.body.length === 0) {
```
lemonmade: Let's move this check into the filter of `find` (`body: {body: (body) => body.length === 0}` in the `handler` clause.

packages/shopify-codemod/transforms/remove-empty-catch-blocks.js
```diff
@@ -0,0 +1,16 @@
+export default function removeEmptyCatchBlocks({source}, {jscodeshift: j}, {printOptions = {}}) {
+
+  return j(source)
+    .find(j.TryStatement, {
+      type: 'TryStatement',
+      handler: {
+        type: 'CatchClause',
```
lemonmade: Can you use `j.TryStatement.name`/ `j.CatchClause.name` instead of the strings? I know we are inconsistent about this but I'd like to stick to that style.

packages/shopify-codemod/transforms/remove-empty-catch-blocks.js
```diff
@@ -0,0 +1,16 @@
...
+      },
+    })
+      .forEach(({node}) => {
```
lemonmade: Indentation a little weird here.

packages/shopify-codemod/transforms/remove-empty-catch-blocks.js
```diff
@@ -0,0 +1,16 @@
...
+        if (node.handler.body.body.length === 0) {
+          delete node.handler;
```
lemonmade: Also, let's try this using the `NodePath` methods — `path.get('handler').replace()`

packages/shopify-codemod/test/fixtures/remove-empty-catch-blocks/basic.input.js
```diff
@@ -0,0 +1,17 @@
...
+  foo();
+} catch (undefined) {
+
+} finally {
+  bar();
+}
+
+try {
+  foo();
+} catch (e) {} finally {
```
lemonmade: Please add a test case with a non-empty `catch` block to show they are not removed.

lemonmade:
Also, remember docs + adding to the `esify` `lib/codemod` manifest.

fandy:
Updated. :)

packages/shopify-codemod/transforms/remove-empty-catch-blocks.js
fandy: In that case I should probably update the other transforms to use `j.Type.name` instead of literals.

lemonmade: There's a lot of those, we can do it later (we can even write a quick transform to do it for us :P)

fandy:
@lemonmade is this PR ready to merge?

packages/shopify-codemod/test/fixtures/remove-empty-catch-blocks/basic.output.js
```diff
@@ -0,0 +1,15 @@
...
+
+try {
+  foo();
+} finally {
+  bar();
+}
+
+try {
+  foo();
+} finally {}
```
lemonmade: We are missing a `try` with a `catch` but without a `finally` — The catch can't be removed in that case.

lemonmade:
One minor comment, otherwise LGTM 👍

GoodForOneFare:
Fixed in decaf.

fandy:
In that case I'll close this branch.

Issue #103, opened 2016-05-23
Clean up `no-cond-assign` errors


GoodForOneFare:
An `if` condition with an assignment should be split into two lines:

```js
// Before
if (img = files[0]) {
  foo();
}

// After
img = files[0];
if (img) {
  foo();
}
```

Other examples from `shopify/shopify` code:
```js
if (href = button.href) {
if (message = $node.text()) {
if (previousInstance = previousContext[key]) {
```

PR #154, opened 2016-06-02
Splits up assignments in if statement parameters

fandy:
Fixes https://github.com/Shopify/javascript/issues/103

/cc @lemonmade


packages/esify/index.js
```diff
@@ -35,6 +33,16 @@ var TRANSFORMS = [
   {path: 'js-codemod/transforms/template-literals'},
   {path: 'shopify-codemod/transforms/strip-template-literal-parenthesis'},
   {path: 'js-codemod/transforms/object-shorthand'},
+<<<<<<< Updated upstream:packages/esify/index.js
```
lemonmade: cf_xczbw4aahquk

packages/shopify-codemod/transforms/split-if-statement-assignment-parameter.js
```diff
@@ -0,0 +1,34 @@
+import {forEachAssignment} from './utils';
+
+export default function splitIfStatementAssignmentParameter({source}, {jscodeshift: j}, {printOptions = {}}) {
+  const sourceAST = j(source);
+  sourceAST
+    .find(j.IfStatement, {
+      type: j.IfStatement.name,
```
lemonmade: Type is a guarantee since you are looking for `j.IfStatement`

packages/shopify-codemod/transforms/split-if-statement-assignment-parameter.js
```diff
@@ -0,0 +1,34 @@
...
+  sourceAST
+    .find(j.IfStatement, {
+      test: {
+        type: j.SequenceExpression.name,
+        expressions: (expressions) => expressions.length > 1,
+      },
+    }).forEach((nodePath) => {
+      for (const params of Array.from(nodePath.node.test.expressions)) {
+        forEachAssignment(params, () => {
+          nodePath.get('test').replace(j.identifier(params.left.name));
```
lemonmade: Could the assignment be something other than an `Identifier`?

packages/shopify-codemod/transforms/split-if-statement-assignment-parameter.js
```diff
@@ -0,0 +1,34 @@
...
+      });
+    });
+  sourceAST
+    .find(j.IfStatement, {
+      test: {
+        type: j.SequenceExpression.name,
+        expressions: (expressions) => expressions.length > 1,
+      },
+    }).forEach((nodePath) => {
+      for (const params of Array.from(nodePath.node.test.expressions)) {
```
lemonmade: I'd prefer something like `nodePath.get('test', 'expressions').forEach` so that we can get the paths of each expression.

packages/shopify-codemod/transforms/split-if-statement-assignment-parameter.js
lemonmade: This is also going to be an issue because you are completely replacing the test with the last expression. This will cause a different behaviour for something that starts out as the following:

```js
if (foo = false, bar = true) { doSomething(); }
```

packages/shopify-codemod/transforms/split-if-statement-assignment-parameter.js
```diff
@@ -0,0 +1,34 @@
...
+        forEachAssignment(params, () => {
```
lemonmade: How does this treat multiple assignment expressions (`if (foo = bar = baz) {}`)?

packages/shopify-codemod/transforms/split-if-statement-assignment-parameter.js
lemonmade: ^ For that, I'd like to see the resulting values be joined with `&&`

packages/shopify-codemod/transforms/split-if-statement-assignment-parameter.js
lemonmade: `params` is also a weird name for this.

packages/shopify-codemod/transforms/utils.js
```diff
@@ -14,6 +14,24 @@ export function findLastMember(node) {
   return node;
 }
 
+export function forEachAssignment(assignment, cb) {
+  if (assignment.right && assignment.left) {
```
lemonmade: When would an assignment not have both a right and a left?

packages/shopify-codemod/transforms/utils.js
```diff
@@ -14,6 +14,24 @@ export function findLastMember(node) {
   return node;
 }
 
+export function forEachAssignment(assignment, cb) {
```
lemonmade: I'm not a huge fan of the helper in its current form. I'd rather see each assignment expression captured and returned as an array, making it call the callback as well feels awkward and out of place with our other helpers.

packages/shopify-codemod/transforms/utils.js
```diff
@@ -14,6 +14,24 @@ export function findLastMember(node) {
...
+  if (assignment.right && assignment.left) {
+    if (j.AssignmentExpression.check(assignment.right)) {
+      let rightName;
+      if (assignment.left.type === 'Identifier') {
+        rightName = j.identifier(assignment.right.left.name);
+      } else {
+        rightName = j.memberExpression(assignment.right.left.object, assignment.right.left.property);
+      }
+      forEachAssignment(assignment.right, cb);
+      return cb(j.assignmentExpression('=', assignment.left, rightName));
```
lemonmade: I would not expect that this helper is creating additional assignment expressions for me.

packages/esify/index.js
```diff
@@ -35,6 +35,13 @@ var TRANSFORMS = [
   {path: 'js-codemod/transforms/template-literals'},
   {path: 'shopify-codemod/transforms/strip-template-literal-parenthesis'},
   {path: 'js-codemod/transforms/object-shorthand'},
+  // These are run very late to ensure they catch any identifiers/ member expressions
+  // added in earlier transforms
+  {path: 'js-codemod/transforms/unquote-properties'},
+  {path: 'shopify-codemod/transforms/computed-literal-keys-to-dot-notation'},
+  {path: 'shopify-codemod/transforms/rename-identifier'},
+  {path: 'shopify-codemod/transforms/rename-property'},
+  {path: 'shopify-codemod/transforms/split-return-assignments'},
```
lemonmade: Were none of these being included before? 😱

lemonmade: Also, docs for the new transform :)

packages/shopify-codemod/transforms/split-if-statement-assignment-parameter.js
lemonmade: Never mind, `&&` will not have the same meaning.

fandy:
Refactored.

The docs will be pushed in a separate PR in the https://github.com/Shopify/javascript/tree/segmented-codemod-docs branch.

What do you think, @lemonmade ?

packages/shopify-codemod/transforms/split-if-statement-assignment-parameter.js
lemonmade: Comment not addressed

packages/shopify-codemod/transforms/split-if-statement-assignment-parameter.js
```diff
@@ -0,0 +1,30 @@
...
+    .find(j.IfStatement, {
+      type: j.IfStatement.name,
+      test: {
+        type: j.AssignmentExpression.name,
+      },
+    }).forEach((nodePath) => {
+      const args = nodePath.get('test').node;
+      nodePath.insertBefore(j.expressionStatement(args));
+      nodePath.get('test').replace(args.left);
+
```
lemonmade: Extra space

packages/shopify-codemod/transforms/split-if-statement-assignment-parameter.js
```diff
@@ -0,0 +1,30 @@
+export default function splitIfStatementAssignmentParameter({source}, {jscodeshift: j}, {printOptions = {}}) {
+  const sourceAST = j(source);
+  sourceAST
+    .find(j.IfStatement, {
+      type: j.IfStatement.name,
+      test: {
+        type: j.AssignmentExpression.name,
+      },
+    }).forEach((nodePath) => {
+      const args = nodePath.get('test').node;
```
lemonmade: You `get` the test node again in a few lines, maybe store it as a temporary variable?

packages/shopify-codemod/transforms/split-if-statement-assignment-parameter.js
```diff
@@ -0,0 +1,30 @@
...
+    });
+  sourceAST
+    .find(j.IfStatement, {
+      test: {
+        type: j.SequenceExpression.name,
+        expressions: (expressions) => expressions.length > 1,
+      },
+    }).forEach((nodePath) => {
+      const args = nodePath.get('test').node;
+      args.expressions = args.expressions.map((assignment) => {
```
lemonmade: I think it would be better to do something like this (if it's possible; you are modifying the expression in a loop so maybe not?)

```js
nodePath.get('test').forEach((expressionPath) => {
  nodePath.insertBefore(j.expressionStatement(assignment));
  expressionPath.replace(expressionPath.get('left').node);
});

File: packages/shopify-codemod/transforms/split-if-statement-assignment-parameter.js [hunk 13]
```diff
@@ -0,0 +1,30 @@
...
+      const args = nodePath.get('test').node;
+      nodePath.insertBefore(j.expressionStatement(args));
+      nodePath.get('test').replace(args.left);
+
+    });
+  sourceAST
+    .find(j.IfStatement, {
+      test: {
+        type: j.SequenceExpression.name,
+        expressions: (expressions) => expressions.length > 1,
```
lemonmade: This will fail for expressions that are not assignment expressions (`if (foo, bar) {}`). Maybe add a test for `If (foo = bar, baz) {}`, and one for `if (foo = bar, baz = qux) {}` and write the code to make those pass.

lemonmade:
A couple small comments, and one big one. I'm cool with the docs being in a separate PR 👍 you'll also need to rebase on latest master to resolve the conflicts.

fandy:
Updated @lemonmade. :)

packages/shopify-codemod/transforms/split-if-statement-assignment-parameter.js
lemonmade: Still here :P

packages/shopify-codemod/transforms/split-if-statement-assignment-parameter.js
lemonmade: Unresolved

packages/shopify-codemod/transforms/split-if-statement-assignment-parameter.js
lemonmade: Unresolved

packages/shopify-codemod/transforms/split-if-statement-assignment-parameter.js
lemonmade: I'd still like to see us work on the path level for this.

packages/shopify-codemod/transforms/split-if-statement-assignment-parameter.js
lemonmade: Pretty sure this is a redundant check now, since any sequence expression probably requires multiple expressions.

packages/shopify-codemod/transforms/utils.js
lemonmade: Is this helper still needed?

GoodForOneFare:
I adapted this into a smaller PR: https://github.com/Shopify/javascript/pull/192.

PR #165, opened 2016-06-13
Add `args` spread argument to functions that reference `arguments`

GoodForOneFare:
Way trickier than it sounds due to IIFE scope quirks.

/cc @lemonmade, @fandy, @justinthec

### DIFF
--- a/packages/shopify-codemod/test/fixtures/arguments-to-args-spread/basic.input.js
+++ b/packages/shopify-codemod/test/fixtures/arguments-to-args-spread/basic.input.js
@@ -1,0 +1,59 @@
+function shouldGetArgsSpread() {
+  foo(arguments);
+}
+
+function shouldGetArgsSpread() {
+  foo(...arguments);
+}
+
+function shouldGetArgsSpread() {
+  console.log(arguments);
+  function arg1ShouldPreventRenamesInsideThisScope(arg1) {
+    console.log(arguments);
+  }
+}
+
+function foo() {
+  (function() { // Function IIFE has its own scope.
+    console.log(arguments);
+  })();
+}
+
+function foo() {
+  console.log(arguments);
+  (() => { // Arrow IIFEs with no arguments inherit parent's arguments.
+    console.log(arguments);
+    (() => {
+      console.log(arguments);
+      (() => {
+        console.log(arguments);
+      })();
+    })();
+  })();
+}
+
+function arrowIIFEParentWithArgsSpread(...args) {
+  (() => { // Arrow IIFE with no arguments inherits parent scope's arguments.
+    console.log(arguments);
+  })();
+}
+
+function foo() {
+  (() => {
+    (() => {
+      console.log(arguments);
+    })("arg1"); // Arrow IIFE with arguments does not inherit parent's arguments.
+  })();
+}
+
+function foo() {
+  console.log(arguments);
+  (() => {
+    console.log(arguments);
+    ((arg1) => { // Arrow IIFE with arguments does not inherit parent's arguments.
+      (() => {
+        console.log(arguments);
+      })();
+    })();
+  })();
+}
--- a/packages/shopify-codemod/test/fixtures/arguments-to-args-spread/basic.input.js
+++ b/packages/shopify-codemod/test/fixtures/arguments-to-args-spread/basic.input.js
@@ -7,53 +7,31 @@ function shouldGetArgsSpread() {
 }
 
 function shouldGetArgsSpread() {
-  console.log(arguments);
-  function arg1ShouldPreventRenamesInsideThisScope(arg1) {
+  foo(() => {
+    console.log(arguments);
+  });
+}
+
+function noArgsSpread() {
+  function shouldGetArgsSpread() {
     console.log(arguments);
   }
 }
 
-function foo() {
-  (function() { // Function IIFE has its own scope.
-    console.log(arguments);
-  })();
-}
-
-function foo() {
-  console.log(arguments);
-  (() => { // Arrow IIFEs with no arguments inherit parent's arguments.
-    console.log(arguments);
-    (() => {
-      console.log(arguments);
-      (() => {
-        console.log(arguments);
-      })();
-    })();
-  })();
-}
-
 function arrowIIFEParentWithArgsSpread(...args) {
-  (() => { // Arrow IIFE with no arguments inherits parent scope's arguments.
+  (() => {
     console.log(arguments);
   })();
 }
 
 function foo() {
-  (() => {
-    (() => {
-      console.log(arguments);
-    })("arg1"); // Arrow IIFE with arguments does not inherit parent's arguments.
-  })();
+  ((iifeArg) => {
+    console.log(iifeArg, arguments);
+  })('iifeArg1');
 }
 
 function foo() {
-  console.log(arguments);
-  (() => {
-    console.log(arguments);
-    ((arg1) => { // Arrow IIFE with arguments does not inherit parent's arguments.
-      (() => {
-        console.log(arguments);
-      })();
-    })();
-  })();
+  ((...bar) => {
+    console.log(bar, arguments);
+  })('barValue');
 }
--- a/packages/shopify-codemod/test/fixtures/arguments-to-args-spread/nested-iife.input.js
+++ b/packages/shopify-codemod/test/fixtures/arguments-to-args-spread/nested-iife.input.js
@@ -1,0 +1,16 @@
+function foo() {
+  (function() { // Function IIFE has its own scope.
+    console.log(arguments, 'a' === args[0]);
+
+    (() => {
+      console.log(arguments, 'a' === args[0]);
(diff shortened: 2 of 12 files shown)


packages/shopify-codemod/transforms/arguments-to-args-spread.js
```diff
@@ -0,0 +1,65 @@
...
+
+  function hasNoArgsVar(path) {
+    return path.scope.lookup('args') == null;
+  }
+
+  const sourceAST = j(source);
+
+  const argumentsReferences = sourceAST
+    .find(j.Identifier, {name: 'arguments'})
+    .filter(hasSpreadableRootScope)
```
fandy: Could you combine `hasNoArgsVar` and `hasSpreadableRootScope` into a single `.filter`?

```
.filter(path => hasNoArgsVar || hasSpreadableRootScope)
```

justinthec: It would look like: `.filter(path => hasNoArgsVar(path) || hasSpreadableRootScope(path))`

I find it harder to read and most of the time function chaining is preferred (easier to read, easier to add and remove from the chain).

GoodForOneFare: @justinthec said it nicely :)  I'm not averse to either version, but I think separate filters is more readable in this context.

packages/shopify-codemod/transforms/arguments-to-args-spread.js
```diff
@@ -0,0 +1,65 @@
+export default function argumentsToArgsSpread({source}, {jscodeshift: j}, {printOptions = {quote: 'single'}}) {
+  function isArglessIIFE(path) {
```
lemonmade: I find this name a little misleading because it's really looking for an IIFE that has no arguments **and** doesn't bind `arguments` (the arrow function part of it).

packages/shopify-codemod/transforms/arguments-to-args-spread.js
```diff
@@ -0,0 +1,65 @@
...
+  function findNonIIFEScope(path) {
+    let scope = path.scope;
+    while (isArglessIIFE(scope.path.parentPath)) {
+      scope = scope.parent;
+    }
+
+    return scope.path;
+  }
+
+  function isFunctionScopeWithNoParams({scope}) {
```
lemonmade: Won't this scope just be the same as the function path that is passed in to it?

packages/shopify-codemod/transforms/arguments-to-args-spread.js
```diff
@@ -0,0 +1,65 @@
...
+    return scope.node.params.length === 0;
+  }
+
+  function isFunctionScopeWithArgsSpread({node: {params}}) {
+    return (
+      j.RestElement.check(params[0]) &&
+      params[0].argument.name === 'args'
```
lemonmade: Worth supporting different names for this?

packages/shopify-codemod/transforms/arguments-to-args-spread.js
```diff
@@ -0,0 +1,65 @@
...
+  const sourceAST = j(source);
+
+  const argumentsReferences = sourceAST
+    .find(j.Identifier, {name: 'arguments'})
+    .filter(hasSpreadableRootScope)
+    .filter(hasNoArgsVar);
+
+  // Push `...args` into root scope param lists.
+  argumentsReferences
+    .map((path) => path.scope.path)
```
lemonmade: Why do you need this part?

packages/shopify-codemod/transforms/arguments-to-args-spread.js
```diff
@@ -0,0 +1,65 @@
...
+    .map(findNonIIFEScope)
+    .filter(isFunctionScopeWithNoParams)
```
lemonmade: Can you just do this regardless of whether it has params or not, and just always replace the params with your rest element?

packages/shopify-codemod/test/fixtures/arguments-to-args-spread/basic.input.js
```diff
@@ -0,0 +1,59 @@
...
+  (() => { // Arrow IIFE with no arguments inherits parent scope's arguments.
+    console.log(arguments);
+  })();
+}
+
+function foo() {
+  (() => {
+    (() => {
+      console.log(arguments);
+    })("arg1"); // Arrow IIFE with arguments does not inherit parent's arguments.
```
lemonmade: Maybe I am misreading this comment, but doesn't an arrow function always have its parent version of arguments? This doesn't seem to make sense given that you are (correctly) transforming this to use `args`, unless I am misreading.

packages/shopify-codemod/test/fixtures/arguments-to-args-spread/sanity.input.js
```diff
@@ -0,0 +1,26 @@
+function argNamesShouldPreventSpread(arg1, arg2) {
```
lemonmade: Could use some non-arrow function IIFE examples to show how they behave differently.

packages/shopify-codemod/transforms/arguments-to-args-spread.js
```diff
@@ -0,0 +1,65 @@
+export default function argumentsToArgsSpread({source}, {jscodeshift: j}, {printOptions = {quote: 'single'}}) {
+  function isArglessIIFE(path) {
+    return j.match(path, {
+      type: j.CallExpression.name,
+      callee: {
+        type: j.ArrowFunctionExpression.name,
+        params: (params) => params.length === 0,
+      },
+      arguments: (args) => args.length === 0,
```
lemonmade: I could be wrong, but I am not sure that it matters if the IIFE is passed any arguments.

GoodForOneFare:
@lemonmade you're right about IIFE arguments not affecting the value of `arguments`.  Not sure what I was doing to make me think otherwise :/

Anyway, I've greatly simplified the PR and added support for reusing existing spread names.

packages/shopify-codemod/test/fixtures/arguments-to-args-spread/nested-iife.input.js
```diff
@@ -0,0 +1,16 @@
...
+  (function() { // Function IIFE has its own scope.
+    console.log(arguments, 'a' === args[0]);
+
+    (() => {
+      console.log(arguments, 'a' === args[0]);
+      (() => {
+        console.log(arguments, 'a' === args[0]);
+
+        (function() {
+          console.log(arguments, 'd' === args[0]);
```
lemonmade: Nitpick, but might make more sense to use `arguments` since `args` is what it will be renamed to after the transform.

GoodForOneFare: Good point.  Done!

packages/shopify-codemod/transforms/arguments-to-args-spread.js
```diff
@@ -0,0 +1,42 @@
+export default function argumentsToArgsSpread({source}, {jscodeshift: j}, {printOptions = {quote: 'single'}}) {
+  function isArrowFunction(path) {
```
lemonmade: This can probably just be `j.ArrowFunctionExpression.check(path.node)` now.

packages/shopify-codemod/transforms/arguments-to-args-spread.js
```diff
@@ -0,0 +1,42 @@
...
+  function hasSpreadableRootScope(path) {
+    const rootScope = findNonArrowScope(path);
+    return (
+      rootScope.node.params.length === 0 ||
+      j.RestElement.check(rootScope.node.params[0])
+    );
+  }
+
+  function hasNoArgsDeclaration({scope}) {
+    return scope.lookup('args') == null;
```
lemonmade: I think there's a `hasBinding` that might be more appropriate here?

packages/shopify-codemod/transforms/arguments-to-args-spread.js
```diff
@@ -0,0 +1,42 @@
...
+    .find(j.Identifier, {name: 'arguments'})
+    .filter(hasSpreadableRootScope)
+    .filter(hasNoArgsDeclaration)
+    .forEach((argumentsPath) => {
+      const rootScope = findNonArrowScope(argumentsPath).node;
+      if (rootScope.params.length === 0) {
+        rootScope.params = [j.restElement(j.identifier('args'))];
+      }
+
+      argumentsPath.node.name = rootScope.params[0].argument.name;
```
lemonmade: I think there might be a rename method on the path? Could be wrong.

lemonmade:
Minor comments, otherwise LGTM 👍

GoodForOneFare:
Now using `scope.declares`, `path.replace`, and `j.ArrowFunctionExpression.check`.  I'll 🚢 when green.

Thank you for always making my code so much better.

PR #168, opened 2016-06-13
Moved all transforms from README to it's own in docs folder

fandy:
Documentation in `packages/shopify-codemod/README.md` doesn't scale as we add more transforms. This PR simply separates them into a docs folder like so `packages/shopify-codemod/docs/some-transform.md`.

**After this PR is merged you should write documentation for new codemods under the new docs folder.**

/cc @lemonmade @GoodForOneFare @justinthec


lemonmade:
👍 Can you add an auto-generated markdown file as part of `bin/create-transform`, and teach `bin/rename-transform` to rename the doc file too?

GoodForOneFare:
There's a risk of this backing things up.  I'll create a separate issue for updating the transforms, and merge this in now.

Issue #158, opened 2016-06-06
Improve ternaries that decaf to IIFEs


lemonmade:
From CoffeeScript, we get many of these gross IIFEs that are basically just ternary expressions:

```js
(() => {
  if (foo) {
    return bar;
  } else {
    return baz;
  }
 })()
```

This pattern (IIFE, containing only a conditional, optionally with an `else` clause, where each branch of the conditional returns some value) should be transformed into something like `foo ? bar : baz`, assuming that the containing expression can take a ternary (`return`, assignment, etc should all be fine to accept a ternary in place of this expression).

justinthec:
Do these not autoclose on PR merge?

GoodForOneFare:
You need "Fixes #158" in the commit message: https://help.github.com/articles/closing-issues-via-commit-messages/

PR #169, opened 2016-06-14
Added IIFE to Ternary transform

justinthec:
https://github.com/Shopify/javascript/issues/158

Turns:
```js
(() => {
  if (foo) {
    return bar;
  } else {
    return baz;
  }
})();
```

into

```js
(() => {
  (foo ? bar : baz);
})();
```

~~If there a way I can just modify the body of a `Node` and then return the original `node`?~~
~~When I try~~
```js
let ternaryExpression = j.conditionalExpression(test, consequentIdentifier, alternateIdentifier);
functionBlockStatement.body = [ternaryExpression];

return node;
```
~~I get this error:~~
```
{test: [object Object], consequent: [object Object], alternate: [object Object], loc: null, type: ConditionalExpression, comments: null} does not match type string
```

~~Also should I be using destructuring assignments for the initial search of nodes?~~

@lemonmade @fandy @GoodForOneFare

### DIFF
--- a/packages/shopify-codemod/test/fixtures/iife-to-ternary-expression/basic.output.js
+++ b/packages/shopify-codemod/test/fixtures/iife-to-ternary-expression/basic.output.js
@@ -1,5 +1,15 @@
 (pass ? bar : baz);
 
+(pass ? bar : baz);
+
+((arg) => {
+  if (pass) {
+    return arg;
+  } else {
+    return baz;
+  }
+})(param);
+
 (() => {
   if (multilineConsequent) {
     bar = 34;
--- a/packages/shopify-codemod/test/fixtures/iife-to-ternary-expression/basic.output.js
+++ b/packages/shopify-codemod/test/fixtures/iife-to-ternary-expression/basic.output.js
@@ -1,0 +1,30 @@
+(pass ? bar : baz);
+
+(() => {
+  if (multilineConsequent) {
+    bar = 34;
+    return bar;
+  } else {
+    return baz;
+  }
+})();
+
+(() => {
+  if (multilineAlternate) {
+    return bar;
+  } else {
+    baz = 34;
+    return baz;
+  }
+})();
+
+(() => {
+  moreThanJustAnIfStatement()
+  if (fail) {
+    return bar;
+  } else {
+    return baz;
+  }
+})();
+
+someFunction();
--- a/packages/shopify-codemod/transforms/iife-to-ternary-expression.js
+++ b/packages/shopify-codemod/transforms/iife-to-ternary-expression.js
@@ -1,7 +1,7 @@
 export default function iifeToTernaryExpression({source}, {jscodeshift: j}, {printOptions = {}}) {
   function isImmediatelyInvokedFunctionExpression(expressionPath) {
     let expression = expressionPath.node.expression
-    return expression.type == 'CallExpression' && expression.callee.type == 'ArrowFunctionExpression'
+    return expression.type === 'CallExpression' && expression.callee.type === 'ArrowFunctionExpression'
   }
 
   function containsOnlyIfStatement(functionNode) {
@@ -14,9 +14,9 @@ export default function iifeToTernaryExpression({source}, {jscodeshift: j}, {pri
   }
 
   function nodeIsBlockStatementContainingOnly(node, childType) {
-    return node.type == 'BlockStatement' &&
-           node.body.length == 1 &&
-         node.body[0].type == childType;
+    return node.type === 'BlockStatement' &&
+           node.body.length === 1 &&
+         node.body[0].type === childType;
   }
 
   function isConvertibleExpression(path) {
--- a/packages/shopify-codemod/transforms/iife-to-ternary-expression.js
+++ b/packages/shopify-codemod/transforms/iife-to-ternary-expression.js
@@ -1,48 +1,36 @@
 export default function iifeToTernaryExpression({source}, {jscodeshift: j}, {printOptions = {}}) {
-  function isImmediatelyInvokedFunctionExpression(expressionPath) {
-    let expression = expressionPath.node.expression
-    return expression.type === 'CallExpression' && expression.callee.type === 'ArrowFunctionExpression'
+  function isIIFE(path) {
+    return j.ArrowFunctionExpression.check(path.node.callee);
   }
 
-  function containsOnlyIfStatement(functionNode) {
-    return nodeIsBlockStatementContainingOnly(functionNode.body, 'IfStatement');
+  function calleeContainsIfStatement(path) {
+    return isBlockStatementContainingOnly(path.get('callee', 'body'), j.IfStatement);
   }
 
-  function onlyReturnsOneOfTwoValues(ifStatementNode) {
-    return nodeIsBlockStatementContainingOnly(ifStatementNode.consequent, 'ReturnStatement') &&
-           nodeIsBlockStatementContainingOnly(ifStatementNode.alternate, 'ReturnStatement');
+  function ifStatementOnlyReturnsTwoValues(path) {
+    const ifStatement = path.get('callee', 'body', 'body', 0);
+    return isBlockStatementContainingOnly(ifStatement.get('consequent'), j.ReturnStatement) &&
+           isBlockStatementContainingOnly(ifStatement.get('alternate'), j.ReturnStatement);
   }
 
-  function nodeIsBlockStatementContainingOnly(node, childType) {
-    return node.type === 'BlockStatement' &&
-           node.body.length === 1 &&
-         node.body[0].type === childType;
+  function isBlockStatementContainingOnly(path, childType) {
+    return j.BlockStatement.check(path.node) &&
+           path.node.body.length === 1 &&
+           childType.check(path.node.body[0]);
   }
 
-  function isConvertibleExpression(path) {
-    return isImmediatelyInvokedFunctionExpression(path) &&
-           containsOnlyIfStatement(path.node.expression.callee) &&
-           onlyReturnsOneOfTwoValues(path.node.expression.callee.body.body[0]);
-  }
+  return j(source)
+    .find(j.CallExpression)
+    .filter(isIIFE)
+    .filter(calleeContainsIfStatement)
+    .filter(ifStatementOnlyReturnsTwoValues)
+    .replaceWith((path) => {
+      const ifStatement = path.get('callee', 'body', 'body', '0');
+      const {test, consequent, alternate} = ifStatement.node;
+      const consequentIdentifier = consequent.body[0].argument;
+      const alternateIdentifier = alternate.body[0].argument;
 
-  const sourceAST = j(source);
-
-  sourceAST
-    .find(j.ExpressionStatement)
-    .filter(isConvertibleExpression)
-    .replaceWith(({node}) => {
-      let {expression: {callee: {body: functionBlockStatement}}} = node;
-    let {body: [{test,
-                   consequent: {body: [{argument: consequentIdentifier}]},
-                   alternate: {body: [{argument: alternateIdentifier}]}
-                  }]} = functionBlockStatement;
-
-      let ternaryExpression = j.conditionalExpression(test, consequentIdentifier, alternateIdentifier);
-      let newFunctionBlockStatement = j.blockStatement([j.expressionStatement(ternaryExpression)]);
-      let newArrowFunction = j.arrowFunctionExpression([], newFunctionBlockStatement, true);
-
-      return j.expressionStatement(j.callExpression(newArrowFunction, []));
-    });
-
-  return sourceAST.toSource(printOptions);
+      return j.conditionalExpression(test, consequentIdentifier, alternateIdentifier);
+    })
+    .toSource(printOptions);
 }
--- a/packages/shopify-codemod/transforms/iife-to-ternary-expression.js
+++ b/packages/shopify-codemod/transforms/iife-to-ternary-expression.js
@@ -1,6 +1,12 @@
 export default function iifeToTernaryExpression({source}, {jscodeshift: j}, {printOptions = {}}) {
-  function isIIFE(path) {
-    return j.ArrowFunctionExpression.check(path.node.callee);
+  function isIIFEWIthNoParams({node}) {
+    return (
+      (
+        j.ArrowFunctionExpression.check(node.callee) ||
+        j.FunctionExpression.check(node.callee)
+      ) &&
+      node.callee.params.length === 0
+    );
   }
 
   function calleeContainsIfStatement(path) {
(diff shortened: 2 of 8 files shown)


packages/shopify-codemod/transforms/iife-to-ternary-expression.js
```diff
@@ -0,0 +1,48 @@
+export default function iifeToTernaryExpression({source}, {jscodeshift: j}, {printOptions = {}}) {
+  function isImmediatelyInvokedFunctionExpression(expressionPath) {
+    let expression = expressionPath.node.expression
+    return expression.type === 'CallExpression' && expression.callee.type === 'ArrowFunctionExpression'
```
GoodForOneFare: When checking types, please use `check` (e.g., `j.CallExpression.check(expression)`).  We're not super consistent about this, but we'll eventually get everything converted to this style.

packages/shopify-codemod/transforms/iife-to-ternary-expression.js
```diff
@@ -0,0 +1,48 @@
+export default function iifeToTernaryExpression({source}, {jscodeshift: j}, {printOptions = {}}) {
+  function isImmediatelyInvokedFunctionExpression(expressionPath) {
+    let expression = expressionPath.node.expression
```
GoodForOneFare: You can use `const` here.  This is probably causing a CI failure.

I recommend adding adding a Git pre-push hook to prevent minor stuff reaching review.  Steps for that: 

* Go to your `shopify/javascript` dir
* `echo 'cd packages/shopify-codemod && npm test' > .git/hooks/pre-push`

This'll run the codemod tests, every time you try to git push, and back out if the tests/linting fail.

packages/shopify-codemod/transforms/iife-to-ternary-expression.js
```diff
@@ -0,0 +1,48 @@
+export default function iifeToTernaryExpression({source}, {jscodeshift: j}, {printOptions = {}}) {
+  function isImmediatelyInvokedFunctionExpression(expressionPath) {
```
GoodForOneFare: I'm all for not using annoying contractions in method names.  It's commendable that you typed all of this out!

That said, `iife` is used in the line immediately above this, and it's a generally accepted abbreviation, so feel free to use it here, too.

packages/shopify-codemod/transforms/iife-to-ternary-expression.js
```diff
@@ -0,0 +1,48 @@
...
+    let expression = expressionPath.node.expression
+    return expression.type === 'CallExpression' && expression.callee.type === 'ArrowFunctionExpression'
+  }
+
+  function containsOnlyIfStatement(functionNode) {
+    return nodeIsBlockStatementContainingOnly(functionNode.body, 'IfStatement');
```
GoodForOneFare: Again, you can avoid using magic strings by passing in `j.IfStatement`, et al as arguments.

packages/shopify-codemod/transforms/iife-to-ternary-expression.js
GoodForOneFare: Oh, and install a linter in whatever text editor you're using.  In Atom, I have `linter-eslint`, and `linter-rubocop`.  Sublime Text has similar packages (which are super annoying to enable, which is why I switched to Atom.)

packages/shopify-codemod/transforms/iife-to-ternary-expression.js
```diff
@@ -0,0 +1,48 @@
...
+  function isConvertibleExpression(path) {
+    return isImmediatelyInvokedFunctionExpression(path) &&
+           containsOnlyIfStatement(path.node.expression.callee) &&
+           onlyReturnsOneOfTwoValues(path.node.expression.callee.body.body[0]);
+  }
+
+  const sourceAST = j(source);
+
+  sourceAST
+    .find(j.ExpressionStatement)
```
GoodForOneFare: Anchoring the search on `ExpressionStatement` will sklip a bunch of transformable code.

This transform should be shooting for stuff like this (actual `shopify/shopify` decaffed code!):
```js
const data = {
  url: (() => {
    if (title) {
      return Shopify.routes.app(this.slug).html;
    } else {
      return null;
    }
  })(),

  name: this.name,
};
```

The result should look something like this:
```js
const data = {
  url: (title ? Shopify.routes.app(this.slug).html : null),
  name: this.name,
};
```

GoodForOneFare: I highly recommending running `esify` at least once on `shopify/shopify`.  It'll give you a feel for the impact you're making on the code (and if you're not, you can say valuable things like, "this transform isn't worth the effort").

Here's the gonzo script that I use to update `esify` to contain my work-in-progress transforms:
```sh
node_version=v5.7.0
global_esify_root="/Users/Gord/.nvm/versions/node/$node_version/lib/node_modules/esify"

cp -R ~/shopify/javascript/packages/shopify-codemod/transforms/** $global_esify_root/node_modules/shopify-codemod/transforms
cp ~/shopify/javascript/packages/esify/bin/esify "$global_esify_root/bin/esify"
cp -R ~/shopify/javascript/packages/esify/lib "$global_esify_root"

rm -fr "$global_esify_root/node_modules/shopify-decaf"
cp -R ~/shopify/decaf/ "$global_esify_root/node_modules/shopify-decaf"
```

packages/shopify-codemod/transforms/iife-to-ternary-expression.js
```diff
@@ -0,0 +1,48 @@
...
+    .filter(isConvertibleExpression)
+    .replaceWith(({node}) => {
+      let {expression: {callee: {body: functionBlockStatement}}} = node;
+    let {body: [{test,
```
GoodForOneFare: I'm not clever enough to unpack what this is doing.  Just from a plain, "what variables are being created here?" perspective.

On top of that, someone trying to understand a transform has to maintain a mental AST, and constantly compare it to the code.  Going crazy with destructuring (to me) moves the code away from a left-to-right traversal that makes relating the AST to code somewhat easy.

I'd go with something more explicit/verbose to help out the enfeebled:
```js
const ifStatement = path.get('expression', 'callee', 'body', 'body', '0');
const {test, consequent, alternate} = ifStatement.node;
```

packages/shopify-codemod/transforms/iife-to-ternary-expression.js
```diff
@@ -0,0 +1,48 @@
...
+                   consequent: {body: [{argument: consequentIdentifier}]},
+                   alternate: {body: [{argument: alternateIdentifier}]}
+                  }]} = functionBlockStatement;
+
+      let ternaryExpression = j.conditionalExpression(test, consequentIdentifier, alternateIdentifier);
+      let newFunctionBlockStatement = j.blockStatement([j.expressionStatement(ternaryExpression)]);
```
GoodForOneFare: Per the sample result in my line 31 comment, there's no need to wrap the ternary in anything.  Once the transform is rooted on `.find(j.CallExpression`, throwing things away should be easy.

packages/shopify-codemod/transforms/iife-to-ternary-expression.js
```diff
@@ -0,0 +1,48 @@
...
+      let newArrowFunction = j.arrowFunctionExpression([], newFunctionBlockStatement, true);
+
+      return j.expressionStatement(j.callExpression(newArrowFunction, []));
+    });
+
+  return sourceAST.toSource(printOptions);
```
GoodForOneFare: Since this is a simple transform, there's no need to store `sourceAST`.  Everything can be chained into once big call.

packages/shopify-codemod/transforms/iife-to-ternary-expression.js
GoodForOneFare: Another example from `shopify/shopify`'s decaffed code:
```js
    this.cents = (() => {
      if (convert) {
        return Shopify.Money._convertToCents(amount);
      } else {
        return amount;
      }
    })();
```

packages/shopify-codemod/transforms/iife-to-ternary-expression.js
```diff
@@ -0,0 +1,48 @@
...
+    return isImmediatelyInvokedFunctionExpression(path) &&
+           containsOnlyIfStatement(path.node.expression.callee) &&
+           onlyReturnsOneOfTwoValues(path.node.expression.callee.body.body[0]);
+  }
+
+  const sourceAST = j(source);
+
+  sourceAST
+    .find(j.ExpressionStatement)
+    .filter(isConvertibleExpression)
```
GoodForOneFare: Trying to avoid compacting things into a vague method name.  As you pointed out in my args spread PR, it's easier to read a few well-named, chained methods than it is to jump around from transform, to method, then back again.

Something like
```js
.filter(isIIFE)
.filter(isCalleeIfStatementWithReturns)
```

packages/shopify-codemod/transforms/iife-to-ternary-expression.js
```diff
@@ -0,0 +1,48 @@
...
+  function containsOnlyIfStatement(functionNode) {
+    return nodeIsBlockStatementContainingOnly(functionNode.body, 'IfStatement');
+  }
+
+  function onlyReturnsOneOfTwoValues(ifStatementNode) {
+    return nodeIsBlockStatementContainingOnly(ifStatementNode.consequent, 'ReturnStatement') &&
+           nodeIsBlockStatementContainingOnly(ifStatementNode.alternate, 'ReturnStatement');
+  }
+
+  function nodeIsBlockStatementContainingOnly(node, childType) {
```
GoodForOneFare: I think we're trying to lean on using path getters (e.g., `path.get('foo', 'bar', 'blah')`).  This simplifies the code by not passing around a mix of path/node refs willy-nilly.

Again, the codebase isn't consistent about this, but prefer paths when you're adding code.

GoodForOneFare:
No tests :(  As I'm not great at running through transforms in my head, I have to copy+paste all of this into astexplorer.  Tests get rid of that review hurdle, get you thinking about edge cases, and make it much easier to refactor your code with confidence.

There are plenty of examples in the project, and you can ping me, Chris, or Andy for a quick intro to writing tests.  It's super easy, yo.

GoodForOneFare:
Remember to add docs for this to the README, and add this to the list of transforms in `esify/lib/codemod.js`.

GoodForOneFare:
This is a really nice first effort!  The if/ternary conversion is sound, so it just has to be applied to a broader scope and we're good to go.

lemonmade:
Good left lots of good comments :) My biggest thing is that I was under the impression we would be getting rid of the IIFE entirely, not just turning the contained conditional into a ternary.

fandy:
This is better than my first transform, and Gord added some great comments.

Some things I wish I knew when I started:

1. Write tests and documentation (Gord, Chris or I) can show you. :)
2. Destructuring is useful, but you should do it in the params only when you don't need access to `nodePath` methods. Otherwise you can destructure in the block statement. Try looking into some [useful `nodePath` methods like `get`, `replace`, and `insertBefore`](https://github.com/facebook/jscodeshift/wiki/jscodeshift-Documentation), and source dig [Jscodeshift](https://github.com/facebook/jscodeshift/tree/master/src) when you're stuck.
3. Don't use literals when adding/ reassigning a node. Instead, use Jscodeshift builder functions:

```
// do:
nodePath.insertAfter(
  j.expressionStatement(
      j.identifier('foo')
  )
)
// don't:
nodePath.insertAfter(
  {
    type: 'expressionStatement',
    expression: {
      type: 'identifier',
      name: 'foo'
    }
  }
)
```

Lastly, in ES6 it's good to default to `const` and switch to `let` only when you need reassigning. A quirk is to `const` is that you can still mutate object props and arrays:

```
// valid
const settings = {
    baseUrl: "https://example.com"
};
settings.baseUrl = "https://evil.example.com";
// valid
const numbers = [1, 2, 3];
numbers.push(4);
numbers.shift();
```

Your linter will remind you of these things once you've set it up: `apm install linter-eslint`

justinthec:
Thanks for all the feedback guys :)

packages/shopify-codemod/transforms/iife-to-ternary-expression.js
```diff
@@ -0,0 +1,36 @@
...
+  function ifStatementOnlyReturnsTwoValues(path) {
+    const ifStatement = path.get('callee', 'body', 'body', 0);
+    return isBlockStatementContainingOnly(ifStatement.get('consequent'), j.ReturnStatement) &&
+           isBlockStatementContainingOnly(ifStatement.get('alternate'), j.ReturnStatement);
+  }
+
+  function isBlockStatementContainingOnly(path, childType) {
+    return j.BlockStatement.check(path.node) &&
+           path.node.body.length === 1 &&
+           childType.check(path.node.body[0]);
```
lemonmade: Restructure in the params to grab `path.node`.

packages/shopify-codemod/transforms/iife-to-ternary-expression.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  function ifStatementOnlyReturnsTwoValues(path) {
+    const ifStatement = path.get('callee', 'body', 'body', 0);
+    return isBlockStatementContainingOnly(ifStatement.get('consequent'), j.ReturnStatement) &&
+           isBlockStatementContainingOnly(ifStatement.get('alternate'), j.ReturnStatement);
+  }
+
+  function isBlockStatementContainingOnly(path, childType) {
+    return j.BlockStatement.check(path.node) &&
```
lemonmade: Personally not a huge fan of the extra-indents-for-lining-up style, I prefer something like

```js
return (
  foo &&
  bar &&
  baz
);
```

when you need to break it up over multiple lines.

packages/shopify-codemod/transforms/iife-to-ternary-expression.js
```diff
@@ -0,0 +1,36 @@
+export default function iifeToTernaryExpression({source}, {jscodeshift: j}, {printOptions = {}}) {
+  function isIIFE(path) {
```
lemonmade: This name is a bit misleading since an IIFE does not need to be an arrow function, it just happens to be for the particular case we are looking at. Does it matter if the IIFE is passed any parameters?

packages/shopify-codemod/test/fixtures/iife-to-ternary-expression/basic.output.js
```diff
@@ -0,0 +1,30 @@
+(pass ? bar : baz);
```
lemonmade: For sanity tests, can you add tests that this transform still works when the IIFE is part of a return statement/ part of an assignment expression?

packages/shopify-codemod/transforms/iife-to-ternary-expression.js
```diff
@@ -0,0 +1,36 @@
...
+  return j(source)
+    .find(j.CallExpression)
+    .filter(isIIFE)
+    .filter(calleeContainsIfStatement)
+    .filter(ifStatementOnlyReturnsTwoValues)
+    .replaceWith((path) => {
+      const ifStatement = path.get('callee', 'body', 'body', '0');
+      const {test, consequent, alternate} = ifStatement.node;
+      const consequentIdentifier = consequent.body[0].argument;
+      const alternateIdentifier = alternate.body[0].argument;
```
lemonmade: Not necessarily identifiers, could be any expressions.

lemonmade:
Minor comments, overall code is very clean and the transform is super easy to read 👍

packages/shopify-codemod/transforms/iife-to-ternary-expression.js
justinthec: It should not matter since the resulting ternary would be share the same scope as the IIFE that it replaces. Any arguments that would be passed in by the caller would also be available to the ternary.

You're right, I'm missing the case of `FunctionExpression`, adding that in.

lemonmade: Not sure this is a real thing, but presumably this would break?

```js
var foo = 1;
var bar = 2;

const baz = ((foo) => {
  if (true) { return foo; } else { return 3; }
})(bar);
```

Your transform will make `baz` `1` instead of `2`. I think it might be worth just bailing out if there are any arguments.

packages/shopify-codemod/test/fixtures/iife-to-ternary-expression/basic.output.js
justinthec: I've added these in `sanity.(input|output).js`

justinthec:
🚢 ?

lemonmade:
👍 Make sure to squash

GoodForOneFare:
Yeah, the filter functions are nicely named in this one 🚢 👍

Issue #116, opened 2016-05-23
Don't pointlessly return in methods that ignore return values


GoodForOneFare:
Some functions like Lodash's `_.each` ignore return values.  To avoid copy+paste anti-patterns, it might be useful to clean those up.

```js
// Before
return _.each(foo, (element, index) => {
  if (index === 0) {
    return; // This return short-circuits following logic, so it has to stay.
  }

  return element.position = index;
});

// After
return _.each(foo, (element, index) => {
  if (index === 0) {
    return; // This return short-circuits following logic, so it has to stay.
  }

  element.position = index;
});
```

Note that `$._each` is influenced by return values, so be sure to ignore those :)

fandy:
Fixed in #171

Issue #115, opened 2016-05-23
Don't return from methods that definitely return `undefined`


GoodForOneFare:
The Coffee=>JS conversion adds a bunch of meaningless return values.  We can reduce copy+paste anti-patterns by removing some of them.  

If they leave behind meaningless returns, we have other transforms to clean that up.

Likely candidates:
* `event.preventDefault()`
* `console.log`
* `console.warn`
* `_.forEach`, `_.each`
* `window.postMessage`

```
// Before
function foo(event) {
  return event.preventDefault();
}

// After
function foo(event) {
  event.preventDefault();
  return;
}
```

fandy:
Fixed in #171

fandy:
Closed the wrong issue :(

PR #171, opened 2016-06-14
Removes pointless returns from specified methods

fandy:
This transform will remove returns in some methods that ignore return values (https://github.com/Shopify/javascript/issues/116).

Docs will be added when https://github.com/Shopify/javascript/pull/168 is merged.

/cc @lemonmade @GoodForOneFare @justinthec


lemonmade:
Thanks for putting this up. I'm going to roll this into a set of transforms that also addresses the (somewhat related) #115 and has a consistent interface for allowing you to pass in the offending object/ value pairs as options.

PR #172, opened 2016-06-14
Allow prototype assignments to exported global

GoodForOneFare:
If `App.Foo = class Foo` is defined, allows in-module assignments to `App.Foo.prototype`.

This clears up the lion's share of `Found multiple exports in a single file, please break up the file first` errors during `esify` (I'll raise a separate issue once I've analyzed the remainder).

It my be possible to fix this in decaf, but I'm putting out this PR because it's already written, and decaf may not fix things 100%.

/cc @lemonmade, @fandy, @justinthec

### DIFF
--- a/packages/shopify-codemod/test/fixtures/global-assignment-to-default-export/local-non-prototype-assignment.input.js
+++ b/packages/shopify-codemod/test/fixtures/global-assignment-to-default-export/local-non-prototype-assignment.input.js
@@ -1,5 +1,0 @@
-App.A = class A {
-
-};
-
-App.A.foo = () => {};
--- a/packages/shopify-codemod/test/fixtures/global-assignment-to-default-export/local-non-prototype-assignment.input.js
+++ b/packages/shopify-codemod/test/fixtures/global-assignment-to-default-export/local-non-prototype-assignment.input.js
@@ -1,0 +1,5 @@
+App.A = class A {
+
+};
+
+App.A.foo = () => {};
--- a/packages/shopify-codemod/transforms/global-assignment-to-default-export.js
+++ b/packages/shopify-codemod/transforms/global-assignment-to-default-export.js
@@ -2,15 +2,11 @@ import {findFirstMember, insertAfterDirectives} from './utils';
 
 export default function globalAssignmentToDefaultExport({source}, {jscodeshift: j}, {printOptions = {}, appGlobalIdentifiers}) {
   function removeFirstMember(memberExpression) {
-    const members = j(memberExpression)
-      .find(j.MemberExpression)
-      .nodes()
-      .slice(-2);
-
-    const last = members.pop();
-    const parent = members.pop() || memberExpression;
-
-    parent.object = last.property;
+    j(memberExpression)
+      .find(j.MemberExpression, {
+        object: (node) => !j.MemberExpression.check(node),
+      })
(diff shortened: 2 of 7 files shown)


packages/shopify-codemod/test/fixtures/global-assignment-to-default-export/local-non-prototype-assignment.input.js
```diff
@@ -0,0 +1,5 @@
+App.A = class A {
+
+};
+
+App.A.foo = () => {};
```
lemonmade: I don't see why we shouldn't transform this one as well (`A.foo = () => {}`)

lemonmade:
One comment on allowing other assignments, but otherwise LGTM 👍

justinthec:
🚢

packages/shopify-codemod/test/fixtures/global-assignment-to-default-export/local-non-prototype-assignment.input.js
GoodForOneFare: Good call.  That clears up a few more transform errors.  Updated, let me know what you think.

packages/shopify-codemod/transforms/global-assignment-to-default-export.js
```diff
@@ -29,10 +47,24 @@ export default function globalAssignmentToDefaultExport({source}, {jscodeshift:
...
+
+            if (!isDefaultExportMemberAssignment(expose, newExpose)) {
+              throw new Error('Found multiple exports in a single file, please break up the file first');
+            }
+
+            const {left, right} = path.node.expression;
+
+            return j.expressionStatement(
+              j.assignmentExpression(
+                '=',
```
lemonmade: I'd probably use the original operator, just to move all of the decision making on what kind of assignment in one spot (the find call).

packages/shopify-codemod/transforms/global-assignment-to-default-export.js
```diff
@@ -1,6 +1,24 @@
 import {findFirstMember, insertAfterDirectives} from './utils';
 
 export default function globalAssignmentToDefaultExport({source}, {jscodeshift: j}, {printOptions = {}, appGlobalIdentifiers}) {
+  function removeFirstMember(memberExpression) {
```
lemonmade: I find this logic kind of confusing. I think you should be able to do something like:

```js
j(memberExpression)
  .find(j.MemberExpression, {
    object: (node) => !j.MemberExpression.check(node)
  })
  .forEach((path) => path.replace(path.node.property));
```

To find the most deeply nested expression (`first.second`) and replace it with just the property (`second`). You should also make sure that `second` is not computed (so, bail out on something like `first[second]`)

GoodForOneFare: 👍 Really nice!  Updated to use that.

The computed `second` scenario won't come up because `isDefaultExportMemberAssignment` is comparing stringified paths that contain a `.`.  If this function was in utils, I'd do it, but I'm going to stay scrappy here.

GoodForOneFare:
The transform will now reuse the original assignment's operator, and I've added a test that verifies non-support for array accessors.

I'll :ship: once things are green.

PR #173, opened 2016-06-14
Add transforms for killing unecessary returns

lemonmade:
Fixes https://github.com/Shopify/javascript/issues/115. Sorry for the big PR, I can break it up if people want.

This PR introduces three transforms: `avoid-returning-unused-results`, `avoid-returning-useless-expressions`, and `remove-unused-expressions`. These collectively solve two cases:

1. Returns in callbacks passed to methods that ignore return values.
2. Returns of method calls that always return `undefined`.

It will also help us get rid of unused expressions separate from the results of these transforms, which is really nice.

cc/ @GoodForOneFare @bouk @justinthec

### DIFF
--- a/packages/esify/README.md
+++ b/packages/esify/README.md
@@ -81,5 +81,44 @@ module.exports = {
       extendWith: 'assignInWith',
     },
   },
+  // A list of object/ property pairs that always ignore return values of any
+  // callbacks passed to them
+  methodsThatIgnoreReturnValues: [
+    {
+      object: '_',
+      methods: ['each'],
+    },
+    {
+      object: /.*/,
+      methods: ['forEach'],
+    },
+  ],
+  // A list of object/ property pairs that always return undefined when called
+  methodsReturningVoid: [
+    {
+      object: 'console',
+      methods: ['log', 'warn'],
+    },
+    {
+      object: /^(e|evt|event)$/,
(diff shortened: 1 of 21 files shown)


packages/esify/README.md
```diff
@@ -81,5 +81,44 @@ module.exports = {
...
+      object: /^(e|evt|event)$/,
+      methods: ['preventDefault'],
+    },
+    {
+      object: 'console',
+      methods: ['log', 'warn'],
+    },
+    {
+      object: /^(e|evt|event)$/,
+      methods: ['preventDefault'],
```
bouk: duplication

packages/shopify-codemod/README.md
```diff
@@ -12,6 +12,83 @@ This repository contains a collection of Codemods written with [JSCodeshift](htt
...
+jscodeshift -t shopify-codemods/transforms/remove-unused-expressions <file>
+```
+
+#### Example
+
+```js
+foo;
+!foo;
+123;
+foo.bar;
```
bouk: technically this one could have a side effect

lemonmade: Yeah, getters can have side effects, but we don't do that anywhere and ESLint will complain about it anyways.

bouk: :+1:

Issue #162, opened 2016-06-13
esify may be causing resource exhaustion


GoodForOneFare:
While running the `esify@12.2.0`, zsh stops functioning with `fork failed: resource temporarily unavailable`, and Atom can't to open new windows.

Eyeballing recent changes, I'm guessing `runESLint` and/or `findESLintBinary` is opening more file descriptors than anticipated.

(Or I could be completely wrong, and I just have to reboot my computer :/ )

GoodForOneFare:
Commented out `runESLint` and things are speedy again.  If the binary can be looked up once then reused, I think things'll improve.

@lemonmade is programmatically calling eslint+fix without `proc.exec` possible?

lemonmade:
Totally possible, I will look into it.

PR #177, opened 2016-06-15
Fix ESLint resource hogging

lemonmade:
Fixes https://github.com/Shopify/javascript/issues/162. Using ESLint's public API to run the lint fixes programatically. Seems to improve things for me!

cc/ @GoodForOneFare


GoodForOneFare:
Started 🎩ing this, and got distracted by Slack.  Completely forgot that it was running, and the end result looks fine 👍 🚢

PR #178, opened 2016-06-15
Remove dash-casing in esify

lemonmade:
cc/ @GoodForOneFare @bouk


GoodForOneFare:
Pragmatism can be frustrating :|  Thanks for taking care of this 👍

Issue #98, opened 2016-05-23
Simplify ?= operator transform output


GoodForOneFare:
`resource_browser.coffee` contains:
```coffeescript
    parentState.__id ?= _.uniqueId()
```

Which becomes the rather verbose:
```js
    if (parentState.__id != null) {
      parentState.__id;
    } else {
      parentState.__id = _.uniqueId();
    }
```

GoodForOneFare:
Still an issue after merging in mainline decaf.  My preference would be to fix it in decaf, but a transform is okay, too.

lemonmade:
Do you know how hard it would be to do this in decaf?

GoodForOneFare:
I take back my decaf preference.  It isn't behaving too badly here, and has to handle edge cases like return-existential-assignment.  Tweaking/adding a transforms would be quicker/easier.

With only decaf running, this Coffee:
```coffeescript
opt["$field"] ?= @element
```

becomes
```js
(opt['$field'] != null ? opt['$field'] : opt['$field'] = this.element);
```

A transform should be able to examine consequents and rewrite if they contain only `Identifier`s/`MemberExpression`s.

lemonmade:
👍 codemod it is!

lemonmade:
@justinthec want to take a crack at this transform next?

GoodForOneFare:
Verified that this is still a problem.

PR #179, opened 2016-06-15
Added ?= decaf output to simple if statement transform

justinthec:
Addresses: https://github.com/Shopify/javascript/issues/98

Turns this `a ?= b` decaf output:
```js
if (a != null) {
  a;
} else {
  a = b;
}
```

into

```js
if (a == null) {
  a = b;
}
```

@lemonmade @GoodForOneFare


packages/shopify-codemod/test/transforms/simplify-?=-if-else-statement.test.js
```diff
@@ -0,0 +1,8 @@
+import 'test-helper';
```
GoodForOneFare: Is there a really, really compelling reason to have `?=` in the file name?  I'd rather not.

justinthec: `conditional-assign-to-if-statement` references `a || a = b` as a `Conditional Assignment`. How would we differentiate between that and `a ?= b`?

GoodForOneFare: 💥  I grant you the power to make up names!  💥  

The CoffeeScript docs hint that it's called `conditional existence operator`, but I don't like that.  Just go with `existential-assignment` until Chris suggests something better ;)

justinthec: Can we actually merge my docs generation so I can use it to rename this? 😂

GoodForOneFare: My main reason for carping about this is that I forget about special symbols when I'm grepping for filenames.  

My incompetence is your inconvenience 👺

packages/shopify-codemod/transforms/simplify-?=-if-else-statement.js
```diff
@@ -0,0 +1,62 @@
...
+
+  function alternateAssignsPrimaryIdentifier(path, primaryIdentifierName) {
+    const alternateExpression = path.get('alternate', 'body', 0, 'expression');
+    return (
+      j.AssignmentExpression.check(alternateExpression.node) &&
+      alternateExpression.get('left', 'name').value === primaryIdentifierName
+    );
+  }
+
+  return j(source)
```
GoodForOneFare: So there's a (totally not documented) structure that our transforms mostly follow:
1. Use `find` to filter by *structure*
2. Use `filter` to filter by *content*
3. Transform with the assumption that 1+2 have taken care of edge cases

I think you could lean more on the first step.  It's commendable that you place great emphasis on creating readable, flowing transforms *but* you can also assume that your reader has a degree of familiarity with `jscodeshift`.

The following is reasonably grokable by someone with a few hours of experience, and is similar in level-of-effort once you account for your eyes being able to stay in one spot.

```
    .find(j.IfStatement, {
      test: {
        type: j.BinaryExpression.name,
        left: {type: j.Identifier.name},
        operator: '!=',
        right: {type: j.Literal.name, value: null},
      },
      consequent: {
        body: (body) => body.length === 1 && j.Identifier.check(body[0].expression),
      },
      alternate: {
        body: (body) => body.length === 1 && j.AssignmentExpression.check(body[0].expression),
      },
    })
    .filter(({node}) => {
      const testId = node.test.left.name;
      const consequentId = node.consequent.body[0].expression.name;
      const alternateId  = node.alternate.body[0].expression.left.name;

      return testId === consequentId && testId === alternateId;
    })
```

GoodForOneFare: And there's probably a middle ground here.  Maybe some helper functions can be reused within the `find` block.  Maybe more concise function names would convince me.  

Take some time to experiment, and see how the variations feel.

packages/shopify-codemod/transforms/simplify-?=-if-else-statement.js
```diff
@@ -0,0 +1,62 @@
...
+    .find(j.IfStatement)
+    .filter(testIsBinaryExpression)
+    .filter(testChecksIdentifierForNotNull)
+    .filter(representsConditionalAssignment)
+    .replaceWith(({node}) => {
+      const {test, alternate} = node;
+      const inverseTest = j.binaryExpression('===', test.left, test.right);
```
GoodForOneFare: We can use `==` here (to catch `undefined`, per https://github.com/Shopify/javascript#11.4).

packages/shopify-codemod/test/transforms/existential-assignment-to-if-statement.test.js
```diff
@@ -0,0 +1,8 @@
+import 'test-helper';
+import existentialAssignmentToIfStatement from 'existential-assignment-to-if-statement';
+
+describe('existentialAssignmentToIfStatement', () => {
+  it('only simplifies if-else statements that have their test, consequent, and alternate all match the coffeescript ?= decaf output', () => {
```
GoodForOneFare: A very long test name (especially with commas) is typically a smell.  A separate `ignores mismatched identifiers` sanity check would be helpful.

GoodForOneFare:
Overall, this is an excellent PR!  The extensive test suite helped me to experiment with my suggestions.

packages/shopify-codemod/test/transforms/simplify-?=-if-else-statement.test.js
lemonmade: 👍 to existential assignment

packages/shopify-codemod/transforms/simplify-?=-if-else-statement.js
```diff
@@ -0,0 +1,62 @@
...
+    );
+  }
+
+  return j(source)
+    .find(j.IfStatement)
+    .filter(testIsBinaryExpression)
+    .filter(testChecksIdentifierForNotNull)
+    .filter(representsConditionalAssignment)
+    .replaceWith(({node}) => {
+      const {test, alternate} = node;
```
lemonmade: Might as well restructure these in the params, too.

packages/shopify-codemod/transforms/simplify-?=-if-else-statement.js
```diff
@@ -0,0 +1,62 @@
+export default function simplifyCondAssignIfElseStatement({source}, {jscodeshift: j}, {printOptions = {quote: 'single'}}) {
+
+  function testIsBinaryExpression(path) {
+    return j.BinaryExpression.check(path.get('test').node);
+  }
+
+  function testChecksIdentifierForNotNull(path) {
+    const test = path.get('test').node;
+    return (
+      j.Identifier.check(test.left) &&
```
lemonmade: This should also be able to work for member expressions. I didn't check what decaf outputs, but CoffeeScript generates the following for `foo.bar.baz ?= qux`:

```js
(foo.bar.baz != null ? foo.bar.baz : foo.bar.baz = qux);
```

(as an aside, the decaf stuff all generates those ternaries — is this codemod done with the assumption that another runs before it to turn those into conditionals?)

justinthec: I was under the assumption that decaf spat out the if statements from the Issue example: https://github.com/Shopify/javascript/issues/98#issue-156184628

justinthec:
Ready for round 2!

Huge thanks to you gord for showing me the powers of `find()` the transform is so small now because of it 🙏

lemonmade:
Awesome job 👍 🚢

packages/shopify-codemod/docs/existential-assignment-to-if-statement.md
```diff
@@ -0,0 +1,23 @@
...
+```js
+if (a != null) {
+  a;
+} else {
+  a = b;
+}
+
+// BECOMES:
+
+if (a === null) {
```
GoodForOneFare: Niggle: this should be `a == null`.

GoodForOneFare:
That was a ton of feedback, and you've nailed all of it 🐑 👍

Issue #176, opened 2016-06-15
Transform create/rename should change documentation filenames


GoodForOneFare:
Now that each transform has its own doc file, `bin/create-transform` and `bin/rename-transform` should handle that.

PR #180, opened 2016-06-15
Add Docs Generation to Helper commands

justinthec:
Addresses: https://github.com/Shopify/javascript/issues/176

When running `bin/create-transform transform-name`:
* `docs/transform-name.md` is created using the template `transform-cli/templates/documentation.md.template`

When running `bin/rename-transform transform-name other-transform-name`:
* `docs/transform-name.md` is renamed to `docs/other-transform-name.md` and has all dash-cased instances of `tranform-name` replaced with `other-transform-name` in its contents.

@GoodForOneFare @lemonmade


GoodForOneFare:
👍 Thanks for taking care of this!  Much appreciated.

PR #186, opened 2016-06-17
Add arrow-body-style guidance for a singular multiline return.

justinthec:
Currently our [`arrow-body-style:as-needed`](http://eslint.org/docs/rules/arrow-body-style#as-needed) rule only allow arrow functions to use block statements when they contain multiple lines.

However when working with an arrow function with a single very long expression, one may be tempted to break it up onto multiple lines using a block statement like so:

```js
body: (body) => {
  return (
    body.length === 1 &&
    (j.Identifier.check(body[0].expression) || j.MemberExpression.check(body[0].expression))
  );
},
```

This however will violate the `arrow-body-style:as-needed` rule.

This added guide to the readme provides the correct alternative to not break the rule while still breaking up the long expression onto multiple lines.

@lemonmade

### DIFF
--- a/README.md
+++ b/README.md
@@ -1353,6 +1353,13 @@ npm run lint
     })
     .filter((x) => {
       return x < 6;
+    })
+    .filter((a) => {
+      return (
+        a &&
+        really &&
+        (long || expression)
+      );
     });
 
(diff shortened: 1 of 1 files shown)


lemonmade:
🐑

Issue #163, opened 2016-06-13
remove-trailing-else-undefined-returns should handle argless returns


GoodForOneFare:
The transform should remove the `else` block here, but it throws an exception.
```js
function foo() {
    if (bar) {
      return bar;
    } else {
      return;
    }
  }
```

Exception
```
TypeError: Cannot read property 'get' of null
    at Function.match 
    at isUndefined (utils.js:33:12)
    at argument (remove-trailing-else-undefined-return.js:13:37)
```

GoodForOneFare:
Fixed by https://github.com/Shopify/javascript/pull/187

PR #187, opened 2016-06-19
Handle argless returns in remove-trailing-else-undefined-return

GoodForOneFare:
/cc @lemonmade, @justinthec


GoodForOneFare:
Fixes #163.

justinthec:
Do you have a test for the case where there is unreachable content after the if-else statement and therefore the return removal would cause it to be reachable?

```js
function foo() {
  if (bar) {
    return bar();
  } else {
    return;
  }
  console.log('unreachable');
}
```

to

```js
function foo() {
  if (bar) {
    return bar();
  }
  console.log('now I am');
}
```

GoodForOneFare:
While I'm reluctant to test edge cases that should never happen in our code, removing dead code would go beyond the intent of this transform.  So I've added a sanity test.  Thanks for the suggestion!

PR #188, opened 2016-06-19
Fix invalid finder criteria in split-return-assignments

### LINKED ISSUE #166: Fix split-return-assignments bugs
For more details, see this comment: https://github.com/Shopify/javascript/commit/bbf5f9c236ceec5c527bbfa708a35c88d7b85914#commitcomment-17838001

The transform should ignore these functions:
```js
function foo() { 
  bar(); // matchLast isn't checking for return statements properly, so it tries to process this.
};

function foo() { 
  return "bar"; // Not sure why `argument: j.AssignmentExpression` is allowing this to match, but it does :(
};
```

GoodForOneFare:
/cc @lemonmade, @justinthec.

Fixed #166.

### DIFF
--- a/packages/shopify-codemod/test/fixtures/split-return-assignments/sanity.input.js
+++ b/packages/shopify-codemod/test/fixtures/split-return-assignments/sanity.input.js
@@ -1,0 +1,7 @@
+function foo() {
+  bar();
+};
+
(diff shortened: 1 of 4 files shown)


justinthec:
Looks good to me 👍

Issue #184, opened 2016-06-16
Import transform should understand named exports


lemonmade:
If we have a file that contains the following:

```js
'expose Foo.Bar'
export function baz() {}
```

And another that contains this CoffeeScript:

```coffee
Foo.Bar.baz();
```

The Coffee file will be converted to:

```js
import Bar from 'path/to/bar';
Bar.baz();
```

Which is wrong because the function is a named export, not a static member of the default export.

lemonmade:
@justinthec would you mind taking a look at this when you get a chance? I think we should be able to add some additional searching to the global-to-import transform to check if it has `export default` and, if not, to do a wildcard import (`import * as Bar from 'path/to/bar'`) instead of a default import.

PR #191, opened 2016-06-20
Modify transform to use namespaced import when finding named exports

justinthec:
Fixes: https://github.com/Shopify/javascript/issues/184

Given a named export:
```js
'expose Foo.Bar'
export function baz() {}
```
Those referencing it will now go from:
```js
Foo.Bar.baz();
```
to using namespace imports:
```js
import * as Bar from 'path/to/bar';
Bar.baz();
```

**Note:** The runtime of the test is marked as red by `npm test` at 106 ms.
image
This is almost certainly due to my use of `fs.fileReadSync`. Any alternative suggestions welcome 🙏 

@lemonmade @GoodForOneFare

### DIFF
--- a/packages/shopify-codemod/transforms/global-reference-to-import.js
+++ b/packages/shopify-codemod/transforms/global-reference-to-import.js
@@ -1,6 +1,7 @@
 import {spawnSync} from 'child_process';
 import {extname, dirname, resolve, relative, join, basename} from 'path';
 import {findFirstMember, findLastMember, insertAfterDirectives} from './utils';
+const fs = require('fs');
 
 function determineFileSearcher() {
   if (spawnSync('which', ['ag']).status === 0) {
@@ -68,6 +69,13 @@ export default function globalReferenceToImport(
     return fileForIdentifier[identifier];
   }
 
+  function searchForNamedExports(filename) {
+    if (filename.endsWith('.coffee'))
+      return false;
+    const namedExports = j(fs.readFileSync(filename).toString())
+      .find(j.ExportNamedDeclaration);
+    return namedExports.paths().length > 0;
+  }
 
   function isGlobalReference(object) {
     return appGlobalIdentifiers.indexOf(findFirstMember(object).name) >= 0;
@@ -101,7 +109,8 @@ export default function globalReferenceToImport(
           const file = getDeclaringFile(member);
           if (file === null) { return null; }
           const name = findLastMember(node).name;
-          imports.set(member, {file, name});
+          const hasNamedExports = searchForNamedExports(file);
+          imports.set(member, {file, name, hasNamedExports});
           return name;
         }
       }
@@ -146,12 +155,17 @@ export default function globalReferenceToImport(
         });
 
       for (const anImport of imports.values()) {
-        const {file, name} = anImport;
+        const {file, name, hasNamedExports} = anImport;
         const {node: {body}} = path;
+        const importSpecifierBuilder = hasNamedExports ? (
+          j.importNamespaceSpecifier
+        ) : (
+          j.importDefaultSpecifier
+        );
         insertAfterDirectives(
           body,
           j.importDeclaration([
-            j.importDefaultSpecifier(j.identifier(name)),
+            importSpecifierBuilder(j.identifier(name)),
           ], j.literal(relativePath(file)))
         );
       }
--- a/packages/shopify-codemod/transforms/global-reference-to-import.js
+++ b/packages/shopify-codemod/transforms/global-reference-to-import.js
@@ -73,6 +73,7 @@ export default function globalReferenceToImport(
     if (extname(filename) === '.coffee') {
       return [];
     }
+    // eslint-disable-next-line no-sync
     const namedExports = j(fs.readFileSync(filename).toString())
       .find(j.ExportNamedDeclaration);
     return namedExports.paths();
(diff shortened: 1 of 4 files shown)


packages/shopify-codemod/transforms/global-reference-to-import.js
```diff
@@ -1,6 +1,7 @@
 import {spawnSync} from 'child_process';
 import {extname, dirname, resolve, relative, join, basename} from 'path';
 import {findFirstMember, findLastMember, insertAfterDirectives} from './utils';
+const fs = require('fs');
```
lemonmade: You can import this just like the other stuff in this file.

packages/shopify-codemod/transforms/global-reference-to-import.js
```diff
@@ -68,6 +69,13 @@ export default function globalReferenceToImport(
     return fileForIdentifier[identifier];
   }
 
+  function searchForNamedExports(filename) {
+    if (filename.endsWith('.coffee'))
+      return false;
```
lemonmade: Linter is probably complaining about this.

packages/shopify-codemod/transforms/global-reference-to-import.js
```diff
@@ -68,6 +69,13 @@ export default function globalReferenceToImport(
...
+    const namedExports = j(fs.readFileSync(filename).toString())
```
lemonmade: I think we can probably get by with a tacky version of this by using the same grepping stuff we use to find the identifier in the first place, but instead of searching for the identifier we just scope our search to only that file with something like `export\s(\S+)`. It's less accurate than doing the AST but should be fine since `export` is a reserved word. Check whether the export is default or not and you should get the same thing. Not sure if this will be faster but it's worth a try.

packages/shopify-codemod/transforms/global-reference-to-import.js
```diff
@@ -146,12 +155,17 @@ export default function globalReferenceToImport(
         });
 
       for (const anImport of imports.values()) {
-        const {file, name} = anImport;
+        const {file, name, hasNamedExports} = anImport;
         const {node: {body}} = path;
+        const importSpecifierBuilder = hasNamedExports ? (
```
lemonmade: If you want to break this onto multiple lines (I don't know that it's necessary in this case anyways), preferred format would be

```js
const importSpecifierBuilder = hasNamedExports
  ? j.importNamespaceSpecifier
  : j.importDefaultSpecifier;
```

I would probably just do this check in the `j.importDeclaration` call instead, though, and store the built identifier so you don't have to repeat `j.identifier(name)`.

packages/shopify-codemod/transforms/global-reference-to-import.js
justinthec: What about cases where the word `export` is used inside a string like `'Click to export products'`. The grep has no feasible way of preventing finds like that.

lemonmade: Do we have ay strings where that's the case? You can always force it to appear at the start of a line, which again, won't catch every case, but might be good enough if it gives us significant speed improvements.

packages/shopify-codemod/transforms/global-reference-to-import.js
justinthec: Something like:
```js
j.importDeclaration([
  (hasNamedExports ? j.importNamespaceSpecifier : j.importDefaultSpecifier)(j.identifier(name)),
],
```
?

That way I don't have to store it, but can use the returned result and immediately invoke it.

lemonmade: It's mostly a matter of preference, but personally I don't love switching on which function to eventually invoke, I find `hasNamedExports ? j.importNamespaceSpecifier(identifier) : j.importDefaultSpecifier(identifier)` cleaner.

packages/shopify-codemod/transforms/global-reference-to-import.js
justinthec: If the transform happens to miss a case due to not using the most accurate way of validating the presence of a named export, wouldn't that result in a wrong import and therefore broken code coming out of the transform?

Why is the performance more important than the integrity of the transform? We only run these transforms once on each file anyways right? Wouldn't we want that 1 time that it's run to be longer but have confidence in the output rather than having the 1 time it's run to be quick but having to double check the result manually to ensure no imports are wrong?

lemonmade: I don't see this as a problem because there is no guarantee that ay of this will work anyways. This code breaks in this case:

```js
'expose Foo.Bar'

export function baz() {}
```

```coffee
Foo.Bar.qux();
```

but we don't really care because that's not a case that actually exists. Similarly, the simple regex might fail for something like this:

```js
'expose Foo.Bar';

const myString = `
export baz;
`; // haha my evil plan has worked!
```

Anyways, just give it a shot, see if there is a meaningful perf difference, and then we can judge whether it's worth it or not.

justinthec:
After testing out the path of using an `ag` search, we found that the performance runtime for the `basic.input.js` test went from `~110ms` to `~150ms` and therefore have decided to stick with the original solution of using `fs` to read the single file.

justinthec:
🚢 ?

lemonmade:
👍

PR #192, opened 2016-06-21
Splits up assignments in if statement parameters

### LINKED ISSUE #103: Clean up `no-cond-assign` errors

GoodForOneFare:
Fixes #103.  Simplified version of Andy's PR: https://github.com/Shopify/javascript/pull/154.

Rebased, cleaned up, and added docs.  Also removed support for `SequenceExpression`s because I don't see those anywhere in the core codebase.

/cc @lemonmade, @justinthec

### DIFF
--- a/packages/esify/lib/codemod.js
+++ b/packages/esify/lib/codemod.js
@@ -3,6 +3,7 @@ var jscodeshift = require('jscodeshift');
 require('babel-register')({ignore: false});
 
 var TRANSFORMS = [
+  {path: 'shopify-codemod/transforms/split-if-assignments'},
   {path: 'shopify-codemod/transforms/coffeescript-soak-to-condition'},
   {path: 'shopify-codemod/transforms/ternary-statement-to-if-statement'},
   {path: 'shopify-codemod/transforms/remove-useless-return-from-test', test: true},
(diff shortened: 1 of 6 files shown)


packages/esify/lib/codemod.js
```diff
@@ -52,6 +53,7 @@ var TRANSFORMS = [
   {path: 'shopify-codemod/transforms/computed-literal-keys-to-dot-notation'},
   {path: 'shopify-codemod/transforms/rename-identifier'},
   {path: 'shopify-codemod/transforms/rename-property'},
+  {path: 'shopify-codemod/transforms/split-return-assignments'},
```
justinthec: Where is this coming from?

packages/shopify-codemod/transforms/split-if-assignments.js
```diff
@@ -0,0 +1,13 @@
+export default function splitIfAssignments({source}, {jscodeshift: j}, {printOptions = {}}) {
+  return j(source)
+    .find(j.IfStatement, {
+      test: {
+        type: j.AssignmentExpression.name,
+      },
+    }).forEach((ifPath) => {
```
justinthec: do we need to name it `ifPath`? I think that `path` is fine since the `find` is explicitly on `j.IfStatement`s

packages/esify/lib/codemod.js
GoodForOneFare: A transform from moons ago.  It was either never added to this list, or was lost in a bad merge.  Either way, it didn't break anything when I added it back, so I left it in here.

packages/shopify-codemod/transforms/split-if-assignments.js
GoodForOneFare: Agreed.  Changed.

justinthec:
This looks very clean 💯

PR #193, opened 2016-06-23
Automatically import Twine for Bindings

bouk:
@lemonmade for after https://github.com/Shopify/shopify/pull/76142

### DIFF
--- a/packages/esify/lib/config.js
+++ b/packages/esify/lib/config.js
@@ -40,6 +40,7 @@ module.exports = function loadConfig() {
       },
       renameIdentifiers: {
         jQuery: '$',
+        Bindings: 'Twine',
       },
       renameProperties: {
         _: {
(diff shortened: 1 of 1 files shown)


lemonmade:
:+1:

packages/esify/lib/config.js
```diff
@@ -35,9 +35,12 @@ module.exports = function loadConfig() {
         NProgress: 'NProgress',
         FastClick: 'shopify-fastclick',
         Clipboard: 'clipboard',
+        Twine: 'twine',
+        Bindings: 'twine',
       },
       renameIdentifiers: {
         jQuery: '$',
+        Bindings: 'Twine',
```
bouk: @lemonmade also added this

lemonmade: @Shopify/tnt are we cool with moving back to `Twine` as the main name for accessing Twine (over `Bindings`)?

tsov: Yes, I think we should 🔥 `Bindings`

lemonmade: 👍

PR #197, opened 2016-06-27
Use a looser peerDependencies for eslint-plugin-shopify

bouk:
@lemonmade we don't want to have to release a new version for every minor eslint upgrade, so let's only pin the major version


lemonmade:
👍

PR #199, opened 2016-06-28
Exports in coffee/js files with the same name are not conflicts

### LINKED ISSUE #198: esify should ignore exports in coffee/js files with the same name
Vestigial code from when esify hyphenated `.js` filenames is causing esify to report erroneous duplicate exports.

GoodForOneFare:
`equivalentJavaScriptFile` was still converting snake=>kebab case.  Also, `.js` paths are now passed into it, so now it returns those without appending an extra `'.js`.

Fixes #198.

/cc @justinthec, @lemonmade.


lemonmade:
👍

Issue #200, opened 2016-06-28
esify should not try to convert .js.erb files into ASTs


GoodForOneFare:
In `global-reference-to-import`, `findDeclaringFile` is returning `.js.erb` files.  `findNamedExports` then attempts to parse those with jscodeshift.

e.g.,:
- `admin/modules/mixins/date_handler.coffee` contains a reference to `Shopify.DateRangeDb`.
- `DateRangeDb` is exposed by `admin/date_range_db.js.erb`

Affects 50+ files in `admin`; probably many more outside that.

GoodForOneFare:
@justinthec the `.js.erb` files are providing useful context, so we can't just ignore them.  I think you created a proof of concept for non-jscodeshift export discovery, so maybe we can just drop that in to fix this?

justinthec:
I could switch the searching strategy based on the file extension. `.js` --> AST, `.js.erb` --> `ag` regex search. Are there any other js file extensions to know about?

GoodForOneFare:
I'd assume that any file type within `app/assets/javascripts` has the potential to show up in search results (sidenote: which probably means its pointlessly sifting through a bunch of `node_modules` folders).

I'm not sure if two code paths for this is worth it.  Was the difference in performance that large?

justinthec:
The difference was about 50 ms

GoodForOneFare:
If the code is already written, then let's go with two code paths.

GoodForOneFare:
Can we get this fixed soon?  It's not going to be a good experience for anyone trying to use esify right now.

justinthec:
I will get the PR up tonight.

PR #203, opened 2016-07-08
Adds regex search fallback for files that are not of a pure .js extension

justinthec:
Fixes: https://github.com/Shopify/javascript/issues/200

Also moves a stray transform `md` file into the `docs` directory.

@GoodForOneFare @lemonmade


packages/shopify-codemod/transforms/global-reference-to-import.js
```diff
@@ -70,13 +70,33 @@ export default function globalReferenceToImport(
...
+
+      const stdout = result.stdout.toString();
+      const exportLines = stdout.trim().split('\n').map(getExportLine);
+      namedExports = filterDefaultExports(exportLines);
     }
-    // eslint-disable-next-line no-sync
-    const namedExports = j(fs.readFileSync(filename).toString())
-      .find(j.ExportNamedDeclaration);
-    return namedExports.paths();
+    }
```
justinthec: There is now a `}` right above another `}` on the same column.

This is due to the collision of 2 `eslint` rules:
 - Switch case indenting requiring `case`/`default` lines to be on the same column as the `switch`.
 - Lexical declarations within `case` statements require a `{ }` wrapper clause to prevent hoisting out of the `case`.

lemonmade: I might just switch to a bunch if `if/else if/ else`, the case stuff is always gross :P

packages/shopify-codemod/transforms/global-reference-to-import.js
```diff
@@ -70,13 +70,33 @@ export default function globalReferenceToImport(
   }
 
   function findNamedExports(filename) {
-    if (extname(filename) === '.coffee') {
-      return [];
+    let namedExports = [];
+    switch (filename.slice(filename.indexOf('.'))) {
```
lemonmade: `path.extname` should give you the same thing.

lemonmade:
Looks good 👍

GoodForOneFare:
👍 Did a quick run of esify with this applied, and things are looking _way_ better.  🚢 once the `case` + `extname` things are fixed, and I'll do a release.

```
Errors:
	'import' and 'export' may only appear at the top level
		app/assets/javascripts/admin/initializers/argus.coffee (6:4)
		app/assets/javascripts/admin/lib/message_bus.coffee (3:2)
		app/assets/javascripts/admin/lib/screen_sharing/helpers.coffee (3:2)
		app/assets/javascripts/admin/lib/screen_sharing/node_store.coffee (3:2)
		app/assets/javascripts/admin/lib/screen_sharing/watcher.coffee (3:2)
	Found multiple exports in a single file, please break up the file first
		app/assets/javascripts/admin/modules/attachment.coffee
		app/assets/javascripts/admin/modules/collections/collection_rules.coffee
		app/assets/javascripts/admin/modules/force_dirty_state_form.coffee
		app/assets/javascripts/admin/modules/funnels/on_demand_deliveries_funnel.coffee
		app/assets/javascripts/admin/modules/home/sales_feature.coffee
		app/assets/javascripts/admin/modules/template_editor.coffee

		app/assets/javascripts/admin/modules/shipping/carrier_calculated_rate.coffee
		app/assets/javascripts/turbo/bar.coffee
	Cannot read property 'value' of null
		app/assets/javascripts/admin/modules/social_marketing_frame.coffee
	Complex binding patterns require an initialization value
		app/assets/javascripts/checkout_next/vendor/jquery.payments.coffee (595:3)
	[12:8] - Illegal use of super() in constructor. super must be called before any this assignments
		test/javascripts/unit/finances/finances_report_test.coffee
	Unexpected token
		test/online-store-javascripts/online_store_test_helper.coffee (221:94)
	Found multiple definitions for Shopify.money_format
		test/storefront-javascripts/unit/option_selection_test.coffee
```

packages/shopify-codemod/transforms/global-reference-to-import.js
justinthec: `extname` will only give `.erb` for `.js.erb`.

After discussing IRL we realized it should be okay since we only care about if the final extension is `.js` or not so I will go back to `extname`.

PR #204, opened 2016-07-13
Minor linting updates

lemonmade:
This PR makes a few small JavaScript linting/ styleguide updates:

- Made it so that `require-flow` allows `@flow` to appear in a line comment (which is apparently allowed now).
- Removed the requirement to omit the braces for arrow functions that have only a return statement and updated the guidance for them to suggest (but not require) omitting the braces for simple return-only arrow functions.
- Turned off linting rules that enforced sorting of properties in React components. This was annoying in my experience and doesn't have the granularity to make it work well with class properties.
- Turned off a rule that required you to use a default export when your module exports only one identifier. IMO some modules make more sense as groups of named exports (usually utility-like things), even if you currently only have one export from that group.

Thoughts? @GoodForOneFare @bouk @tessalt @Shopify/fed

### DIFF
--- a/README.md
+++ b/README.md
@@ -1345,7 +1345,7 @@ npm run lint
 
   ```js
   // bad
-  [1, 2, 3]
+  const result = [1, 2, 3]
     .map((x) => {
       return (x * x) + 1;
     })
@@ -1357,7 +1357,9 @@ npm run lint
   runCallback((foo) => {foo});
 
   // good
-  [1, 2, 3].map((x) => (x * x) + 1).filter((x) => x < 6);
+  const result = [1, 2, 3]
+    .map((x) => (x * x) + 1)
+    .filter((x) => x < 6);
 
   runCallback((foo) => {
     return {foo};
--- a/README.md
+++ b/README.md
@@ -1339,9 +1339,7 @@ npm run lint
   (good) => 'fabulous'
   ```
 
-- [10.1](#10.1) <a name="10.1"></a> If the body of your arrow function takes only a single expression, omit the braces and make use of the implicit `return`. If you are returning an object, however, you must use braces and an explicit `return`.
-
-  > Why? It reads better when multiple such functions are chained together.
+- [10.1](#10.1) <a name="10.1"></a> If the body of your arrow function takes only a single expression, you can omit the braces and make use of the implicit `return`. If your function’s body has more than one expression, it returns an object, or it returns something that spans multiple lines, you can use braces and an explicit `return`.
 
   ESLint rule: [`arrow-body-style`](http://eslint.org/docs/rules/arrow-body-style.html)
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -1353,27 +1351,13 @@ npm run lint
...
   // good
-  [1, 2, 3]
-    .map((x) => (x * x) + 1)
-    .filter((x) => x < 6)
-    .filter((a) => (
-      a &&
-      really &&
-      (long || expression)
-    ));
+  [1, 2, 3].map((x) => (x * x) + 1).filter((x) => x < 6);
```
GoodForOneFare: Should `.filter` be on a new line?

packages/eslint-plugin-shopify/lib/config/rules/ecmascript-6.js
```diff
@@ -2,19 +2,19 @@
 
 module.exports = {
   // Require braces in arrow function body
-  'arrow-body-style': ['error', 'as-needed'],
+  'arrow-body-style': 'off',
```
tessalt: 😂

tessalt: I actually meant that as tears of joy, not laughter.

GoodForOneFare:
👍 LGTM.

packages/eslint-plugin-shopify/lib/config/rules/react.js
```diff
@@ -76,19 +76,7 @@ module.exports = {
...
-        'static-methods',
-        'statics',
-        'constructor',
-        'lifecycle',
-        'everything-else',
-        'render',
-      ],
-    },
-  ],
+  'react/sort-comp': 'off',
```
bouk: :burn: this rule

packages/eslint-plugin-shopify/lib/config/rules/stylistic-issues.js
```diff
@@ -50,7 +50,7 @@ module.exports = {
   // Specify the maximum number of statements allowed per line
   'max-statements-per-line': ['warn', {max: 2}],
   // Require a capital letter for constructors
-  'new-cap': 'off',
+  'new-cap': ['error', {newIsCap: true, capIsNew: false}],
```
bouk: why did you change this but then still disable is in `esnext.js`?

lemonmade: Before now, if you were using the `es5` config, you wouldn't get any version of this rule at all (since the babel plugin is only brought in for `esnext`).

tessalt:
can we get this released?

lemonmade:
@tessalt 13.0.0 released with all the linting changes, let me know if you have nay issues!

PR #205, opened 2016-07-15
Update dependencies

lemonmade:
This PR updates all of our ESLint-related dependencies, including bumping ESLint to 3.x. It also adds a test to make sure that we configure every available rule, and adds some missing rules as a result of that. Here are some highlights:

- I reconfigured the "use stateless React components if you can" rule. I don't think there's a reason not to and it's a very good practice I'd like to encourage.
- Added a bunch of a11y-related ESLint rules
- Allowed constant conditions for loops, so you can do `while (true) {}`

Thoughts? @bouk @GoodForOneFare @Shopify/fed


packages/eslint-plugin-shopify/lib/config/rules/flowtype.js
```diff
@@ -5,6 +5,8 @@ module.exports = {
   'flowtype/require-parameter-type': 'warn',
   // Requires that functions have return type annotation.
   'flowtype/require-return-type': ['warn', 'always', {annotateUndefined: 'never'}],
+  // Makes sure that files have a valid @flow annotation.
+  'flowtype/require-valid-file-annotation': 'warn',
```
bouk: can we delete our plugin now that this exists?

lemonmade: Good call. I think I am also going to turn off the required types for return/ parameter, generally I think inference is better and less overbearing. Let me know if you think that's a bad idea.

packages/eslint-plugin-shopify/lib/config/rules/promise.js
```diff
@@ -1,10 +1,13 @@
...
 module.exports = {
+  'promise/always-catch': 'off',
   // Ensure that each time a then() is applied to a promise, a catch() is applied as well. Exceptions are made if you are returning that promise.
   'promise/catch-or-return': 'warn',
   // Ensure that inside a then() you make sure to return a new promise or value.
   'promise/always-return': 'error',
   // Enforce standard parameter names for Promise constructors.
   'promise/param-names': 'warn',
+  // Ensure that Promise is included fresh in each file instead of relying on the existence of a native promise implementation.
+  'promise/no-native': 'off',
```
tessalt: wait what?

tessalt: nvm, `off` sorry

PR #206, opened 2016-07-18
Testing guide

AWaselnuk:
Hey team. I got the ball rolling with a possible outline for a JS testing guide. Who can iterate on this branch? @tylerball @lemonmade @tessalt @m-ux 

I suggest we a) Get a first version complete before bringing in more FED for review and b) Focus on just writing down what we currently do most of the time rather than advancing the state of the art.

This doc can probably provide a great starting point and I've already lifted some stuff from it: https://github.com/Shopify/shopify/wiki/Admin-Testing

### DIFF
--- a/testing/README.md
+++ b/testing/README.md
@@ -1,0 +1,33 @@
+# JavaScript testing styleguide
+
+This guide provides guidelines for testing JavaScript code.
+Included are recommended tools, approaches to testing, and a living styleguide.
+
+## Table of contents
+
+1. [Approach and levels of testing](#approach-and-leves-of-testing)
+1. [Tooling](#tooling)
+  1. [Tools for a rails stack](#tools-for-a-rails-stack)
+  1. [Tools for a non-rails stack](#tools-for-a-non-rails-stack)
+1. [Assert-style syntax](#assert-style-syntax)
+1. [Spying, stubbing, and mocking](#spying-stubbing-and-mocking)
+1. [Testing React](#testing-react)
+1. [Resources](#resources)
+
+## Approach and levels of testing
+
+## Tooling
+
+### Tools for a rails stack
+
+### Tools for a non-rails stack
+
+## Assert-style syntax
+
+## Spying, stubbin, and mocking
+
+## Testing React
+
+## Resources
+
+[↑ scrollTo('#table-of-contents')](#table-of-contents)
--- a/testing/README.md
+++ b/testing/README.md
@@ -1,25 +1,328 @@
-# JavaScript testing styleguide
+# `import {Testing} from 'Shopify'`
 
-This guide provides guidelines for testing JavaScript code.
-Included are recommended tools, approaches to testing, and a living styleguide.
+This guide looks at our recommended tools and conventions for testing JavaScript at Shopify. The primary focus of this guide is unit testing, but you can find some discussion of other types of testing in the [types of testing section](#types-of-testing).
 
 ## Table of contents
 
-1. [Approach and types of testing](#approach-and-types-of-testing)
-1. [Tooling](#tooling)
-  1. [Tools for a rails stack](#tools-for-a-rails-stack)
-  1. [Tools for a non-rails stack](#tools-for-a-non-rails-stack)
-1. [Assert-style syntax rules](#assert-style-syntax-rules)
-1. [Using fixtures to represent HTML](#using-fixtures-to-represent-html)
-1. [Spying, stubbing, and mocking](#spying-stubbing-and-mocking)
-1. [Testing React](#testing-react)
+1. [Tools](#tools)
+1. [Style](#style)
+1. [Best practices](#best-practices)
+1. [Types of testing](#types-of-testing)
+1. [React](#react)
 1. [Resources](#resources)
+
+
+
+## Tools
+
+We recommend using the following tools in your tests:
+
+- [**mocha**](https://mochajs.org/) is our preferred test framework. It provides the basic scaffolding for tests including hooks, excellent support for asynchronous code, and more.
+
+- [**chai**](http://chaijs.com/) is our preferred assertion library. It provides many expressive assertions that have descriptive error messages in the case of test failures.
+
+- [**sinon**](http://sinonjs.org/) is our preferred library for handling spies and mocks in JavaScript.
+
+These tools are commonly used together and have many excellent online resources, including solid documentation. In addition to the above, you can use [chai plugins](http://chaijs.com/plugins/) to provide more meaningful assertions. Use caution, however: only include a plugin where it materially helps improve the legibility of tests (otherwise, the burden of other developers tracking down a particular assertion outweighs the benefits you get from having it). [chai-as-promised](http://chaijs.com/plugins/chai-as-promised/) is an excellent example of a chai plugin that makes otherwise awkward tests much easier to read.
+
+In addition to the testing framework, you generally need a test *runner*. Which runner to use will depend on the nature of your project:
+
+- For pure-JavaScript projects that do not need access to the DOM, avoid using a test runner and simply run the `mocha` CLI instead (if you need a simple DOM without actual rendering, you can also use [cheerio](https://github.com/cheeriojs/cheerio) to construct a fake DOM).
+
+- For Non-Rails projects that require a full DOM, use [Karma](https://karma-runner.github.io/1.0/index.html). It integrates well with other tools you are likely using, such as Webpack and code coverage tools.
+
+- Rails projects should use [Teaspoon](https://github.com/modeset/teaspoon), which provides an excellent GUI for running tests in the browser. If you are also using [sprockets-commoner](https://github.com/Shopify/sprockets-commoner) to transpile your JavaScript, make sure to include [teaspoon-bundle](https://github.com/Shopify/teaspoon-bundle) to generate your test bundle.
+
+There are a variety of other tools you might need depending on your project. Here are some additional recommendations we make:
+
+- **Test coverage**: we recommend using [nyc](https://github.com/istanbuljs/nyc).
+
+- **Rails fixtures**: if you are testing a component that has a non-trivial HTML component, using the actual partial improves the test’s resilience. We recommend using [Magic Lamp](https://github.com/crismali/magic_lamp) if you need access to real partials in a Rails application (if you are writing a JavaScript application, you should be able to import your partials using JavaScript imports).
+
+[↑ scrollTo('#table-of-contents')](#table-of-contents)
+
+
+
+## Style
+
+- Our [shared linting configuration](../packages/eslint-plugin-shopify) provides a mocha-/ sinon-/ chai-focused linting configuration which you can use to prevent many common mistakes. To use it, simply extend `plugin:shopify/mocha` in your ESLint configuration (full details can be found on that package’s README).
+
+- Use [TDD-style `assert` syntax](http://chaijs.com/api/assert/) over BDD-style `expect` or `should`.
+
+  > Why? This keeps JavaScript tests more in line with our style for writing Ruby tests, which helps everyone contribute to the test codebase more easily.
+
+  ```js
+  // bad
+  expect(2 === 1).to.be.false;
+  (2 === 1).should.be.false;
+
+  // good
+  assert.isFalse(2 === 1);
+  ```
+
+- Use the Mocha-provided [TDD interface](https://mochajs.org/#tdd), which includes the `suite`, `test`, `setup`, and `teardown` functions.
+
+  ```js
+  // bad
+  describe('myFunction', () => {
+    beforeEach(() => {
+      doSomeSetup();
+    });
+
+    it('is a function', () => {
+      assert.isFunction(myFunction);
+    });
+  });
+
+  // good
+  suite('myFunction', () => {
+    setup(() => {
+      doSomeSetup();
+    });
+
+    test('it is a function', () => {
+      assert.isFunction(myFunction);
+    });
+  });
+  ```
+
+- Do not attach variables under test to the Mocha test context (`this`). Instead, use variables scoped to the nearest `suite`. If using ESNext, always use arrow functions to ensure that you will not be able to rely on context.
+
+  > Why? Using a shared context makes it too easy for tests to "leak" (have some impact on other tests), especially in other test files. It also makes it harder to track down what should be in scope at any given time.
+
+  ```js
+  // bad
+  suite('MyComponent', function() {
+    setup(function() {
+      this.subject = new MyComponent();
+      this.doSomethingWithComponent = function() {
+        return this.subject.doSomething();
+      }
+    });
+
+    test(function() {
+      assert.isFalse(this.doSomethingWithComponent());
+    });
+  });
+
+  // good
+  suite('MyComponent', () => {
+    let subject;
+    function doSomethingWithComponent() {
+      return subject.doSomething();
+    }
+
+    setup(() => {
+      subject = new MyComponent();
+    });
+
+    test(() => {
+      assert.isFalse(doSomethingWithComponent());
+    });
+  });
+  ```
+
+- Move any common setup to the `setup` hook of your suite.
+
+  > Why? Setup in the test itself draws attention from whatever is being tested.
+
+  ```js
+  suite('MyComponent', () => {
+    // bad
+    suite('#myBadFunc', () => {
+      test('it does something', () => {
+        const subject = new MyComponent();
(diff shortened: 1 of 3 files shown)


testing/README.md
```diff
@@ -0,0 +1,88 @@
...
+1. [Using fixtures to represent HTML](#using-fixtures-to-represent-html)
+1. [Spying, stubbing, and mocking](#spying-stubbing-and-mocking)
+1. [Testing React](#testing-react)
+1. [Resources](#resources)
+
+## Approach and types of testing
+
+Like all software testing approaches, JavaScript testing can be divided into three types: unit testing, integration testing, and functional testing.
+
+**Unit Testing** Unit testing checks whether a particular piece (unit/module) of code is working fine. The ‘unit’ is often delimited by a JavaScript class.
```
tylerball: In my experience the `unit` in unit testing usually refers to a single method in a class, not the entire class.

testing/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+**Integration Testing** Integration testing is performed when various units are integrated with each other to form a sub-system or a system. This mostly focuses in the design and construction of the software architecture.
+
+**Functional Testing** The software is tested for the functional requirements. This checks whether the application is behaving according to the specification.
+
+We currently have no recommended approach to integration or functional testing. But here are some tips to keep in mind:
+
+* Avoid assertions on CSS selectors as much as possible. These types of tests are often too brittle.
```
tylerball: I'm not sure about this. If your application code refers to CSS selectors, then you should attempt to test them somehow.

lemonmade: I think this was probably in reference to trying to use CSS selectors in integration tests.

tylerball: Oh whoops, of course.

testing/README.md
```diff
@@ -0,0 +1,355 @@
...
+This guide looks at our recommended tools and conventions for testing JavaScript at Shopify. The primary focus of this guide is unit testing, but you can find some discussion of other types of testing in the [types of testing section](#types-of-testing).
+
+## Table of contents
+
+1. [Tools](#tools)
+1. [Style](#style)
+1. [Best practices](#best-practices)
+1. [Types of testing](#types-of-testing)
+1. [React](#react)
+1. [Resources](#resources)
```
lemonmade: I tried to simplify the structure of the guide a bit. Only part I'm not super sure about was splitting style (stylistic rules, generally something we could lint for) and best practices (more overarching issues that affect tests pretty broadly).

testing/README.md
```diff
@@ -0,0 +1,355 @@
...
+
+In addition to the testing framework, you generally need a test *runner*. Which runner to use will depend on the nature of your project:
+
+- For pure-JavaScript projects that do not need access to the DOM, avoid using a test runner and simply run the `mocha` CLI instead (if you need a simple DOM without actual rendering, you can also use [cheerio](https://github.com/cheeriojs/cheerio) to construct a fake DOM).
+
+- For Non-Rails projects that require a full DOM, use [Karma](https://karma-runner.github.io/1.0/index.html). It integrates well with other tools you are likely using, such as Webpack and code coverage tools.
```
lemonmade: I know there's disagreement on the Karma vs Testem thing, anyone want to weigh in? cc/ @tessalt @minasmart

testing/README.md
```diff
@@ -0,0 +1,355 @@
...
+
+- Rails projects should use [Teaspoon](https://github.com/modeset/teaspoon), which provides an excellent GUI for running tests in the browser. If you are also using [sprockets-commoner](https://github.com/Shopify/sprockets-commoner) to transpile your JavaScript, make sure to include [teaspoon-bundle](https://github.com/Shopify/teaspoon-bundle) to generate your test bundle.
+
+There are a variety of other tools you might need depending on your project. Here are some additional recommendations we make:
+
+- **Test coverage**: we recommend using [nyc](https://github.com/istanbuljs/nyc).
```
lemonmade: I've used this pretty successfully in the past.

testing/README.md
```diff
@@ -0,0 +1,355 @@
...
+      });
+
+      test('it does something else', () => {
+        assert.false(subject.doSomethingElse())
+      });
+    });
+  });
+  ```
+
+- Something something about how to name your tests
```
lemonmade: TODO: How are tests named? I am used to using `it()`/ `describe()`/ `context()` where IMO the test naming makes more sense, not sure exactly what to use as the strings for `suite()` and `test()`.

testing/README.md
```diff
@@ -0,0 +1,355 @@
...
+
+[↑ scrollTo('#table-of-contents')](#table-of-contents)
+
+
+
+## Best practices
+
+- Try to limit the number of assertions made in a single test. For unit tests, you should only be testing one thing at a time (generally, an individual function, either standalone or as a method of an object), and more than one or two assertions is generally indicative that you should split your test into multiple tests.
+
+- Something something about how to organize suites (by method? by context?)
```
lemonmade: Similar to question above — do you use nested suites for describing methods? For describing different arguments? Not at all?

testing/README.md
```diff
@@ -0,0 +1,355 @@
...
+
+- In general, each test has three steps: setup, perform the action under test, and assert based on the result.
+
+- Don't rely too heavily on mocking — maybe dependency injection instead?
+
+- Avoid using mocks.
```
lemonmade: Maybe controversial?

testing/README.md
```diff
@@ -0,0 +1,355 @@
...
+
+  global.assert = assert;
+  global.sinon = sandbox;
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+  ```
+
+- Use mocha’s ability to automatically finish tests when a returned promise is resolved for tests that are asynchronous and depend on promises.
```
lemonmade: Maybe controversial?

testing/README.md
```diff
@@ -0,0 +1,355 @@
...
+  // good
+  suite('myFunction', () => {
+    setup(() => {
+      doSomeSetup();
+    });
+
+    test('it is a function', () => {
+      assert.isFunction(myFunction);
+    });
+  });
```
m-ux: cc: @tessalt (given that you just rewrote a bunch of tests using `describe/it`)

testing/README.md
m-ux: 👍  to this question.  @tylerball any insight here?  IIRC you're more used to using TDD style testing than we are.

testing/README.md
```diff
@@ -0,0 +1,355 @@
...
+
+**Functional Testing** The software is tested for the functional requirements. This checks whether the application is behaving according to the specification.
+
+We currently have no recommended approach to integration or functional testing. But here are some tips to keep in mind:
+
+* Avoid assertions on CSS selectors as much as possible. These types of tests are often too brittle.
+* Keep interaction elements like "click button" as open to change as possible.
+* Test more complex flows, for example:
+  * "Add a product" as opposed to the specifics of automatic variant generation or publish visibility toggling.
+
```
m-ux: can we add UI testing to this list as well?

Selenium or an alternative is also a key piece to this equation IMO.

testing/README.md
m-ux: I like this because it reads more clearly, and doesn't depend on any library methods...

testing/README.md
m-ux: can we expand on this one a bit?  I’m not against it, but would like to have more information available about why.

testing/README.md
tessalt: i didn't know mocha had a tdd interface. we used describe/it on brochure

testing/README.md
```diff
@@ -0,0 +1,390 @@
...
+The guide above is mostly focused on unit tests, where a "unit" of code (ideally, the smallest version of an API) is tested in isolation from the rest of the system. However, there are a few other types of testing that are important to consider when writing JavaScript: integration testing, functional testing, and UI testing.
+
+### Integration testing
+
+Integration tests ensure that various parts of a system work properly together. It mostly focuses on the design of an overall architecture rather than individual pieces. In writing integration tests, make sure to keep the following in mind:
+
+- Be careful not to do unit tests for the pieces of the system and integration tests for the system as a whole if the pieces are not part of the public interface. That is, if the pieces will only be used in combination as part of this system, the system is the actual unit of tests, and the pieces are an implementation detail that does not need direct tests (they are, of course, indirectly tested by the testing of the public interface to the system).
+
+### Functional testing
```
lemonmade: @m-ux do you maybe want to add some more notes about browser/ UI tests here? I don't have a ton of experience with them.

lemonmade:
@Shopify/javascript I made some more updates to the guide and would love some feedback from the rest of the group. Are there parts of this that we aren't sold on? Would love to get something out there that we can start pointing people to.

tylerball:
I've only skimmed a bit but this is looking really good so far. I'm working on some paragraphs for an introduction to insert before we start talking about tools, I'd like to address the "Why should I test my Javascript?" question right away.

testing/README.md
tessalt: i'm ok with karma going forward but I'd rather not change from testem on my current project.

testing/README.md
tessalt: does this change error reporting?

testing/README.md
```diff
@@ -0,0 +1,390 @@
...
+
+In the context of developing web applications, functional testing is more accurately described as browser testing. The goal with these tests is to ensure that a part of the application works according to the specifications you have set. Keep the following in mind for these types of tests:
+
+- Avoid relying on class names for finding components on the page. CSS selectors are meant as styling hooks, and the visual appearance of the page should be able to change without breaking your functional tests.
```
tessalt: can we suggest an alternative then?

testing/README.md
```diff
@@ -0,0 +1,390 @@
...
+- Use the most meaningful assertion for what you are trying to test. For a list of available assertions, make sure to check out [chai’s `assert` docs](http://chaijs.com/api/assert/).
+
+  > Why? More specific assertions have more specific error messages, making it easier for you (and any developer that deals with this code in the future) to track down and fix test failures.
+
+  ```js
+  // bad
+  assert.isTrue(foo === bar);
+
+  // good
+  assert.equal(foo, bar);
```
nathanmarks: What about preferring `assert.strictEqual` to avoid type related issues? You can use `assert.ok` and `assert.notOk` for testing truthy/falsy.

testing/README.md
```diff
@@ -0,0 +1,390 @@
...
+  ```js
+  // somewhere in the test setup...
+  // (see http://sinonjs.org/docs/#assert-expose)
+  sinon.assert.expose(assert);
+
+  // bad
+  assert.isTrue(spy.called);
+
+  // good
+  assert.called(spy);
```
nathanmarks: Another good option is testing `callCount`, this can help catch situations where functions are accidentally invoked multiple times.

testing/README.md
```diff
@@ -0,0 +1,390 @@
...
+
+- [**sinon**](http://sinonjs.org/) is our preferred library for handling spies and mocks in JavaScript.
+
+These tools are commonly used together and have many excellent online resources, including solid documentation. In addition to the above, you can use [chai plugins](http://chaijs.com/plugins/) to provide more meaningful assertions. Use caution, however: only include a plugin where it materially helps improve the legibility of tests (otherwise, the burden of other developers tracking down a particular assertion outweighs the benefits you get from having it). [chai-as-promised](http://chaijs.com/plugins/chai-as-promised/) is an excellent example of a chai plugin that makes otherwise awkward tests much easier to read.
+
+In addition to the testing framework, you generally need a test *runner*. Which runner to use will depend on the nature of your project:
+
+- For pure-JavaScript projects that do not need access to the DOM, avoid using a test runner and simply run the `mocha` CLI instead (if you need a simple DOM without actual rendering, you can also use [cheerio](https://github.com/cheeriojs/cheerio) to construct a fake DOM).
```
mikkoh: How does cheerio compare to https://www.npmjs.com/package/jsdom ? I've used `jsdom` in the past and it was REALLY good. I was using it to test React components.

testing/README.md
mikkoh: I'm not we should explicitly mention Webpack since there seems to be a competition between bundlers heating up. I know ppl who are absolutely opposed to things that Webpack does such as overloading `require`. Might be better just to say "bundlers and code coverage tools".

testing/README.md
```diff
@@ -0,0 +1,390 @@
...
+
+  > Why? This keeps JavaScript tests more in line with our style for writing Ruby tests, which helps everyone contribute to the test codebase more easily.
+
+  ```js
+  // bad
+  expect(2 === 1).to.be.false;
+  (2 === 1).should.be.false;
+
+  // good
+  assert.isFalse(2 === 1);
```
mikkoh: If we want to drive home the fact that we want everyone to use Chai we should add in an import for assert because the first time I was skimming through this doc I totally thought this was Node's assert. Using `isFalse` may not be explicit enough to identify this is coming from Chai

testing/README.md
lemonmade: I don't feel too strongly about this, but I've had some issues with JSDom in the past and Cheerio seems to be the popular alternative these days.

testing/README.md
lemonmade: Not that I know of, I believe uncaught rejections will be treated as failures with the error message printed out.

testing/README.md
m-ux: on the cheerio homepage, it _claims_ to be _faster_ than jsdom
> ϟ Blazingly fast: Cheerio works with a very simple, consistent DOM model. As a result parsing, manipulating, and rendering are incredibly efficient. Preliminary end-to-end benchmarks suggest that cheerio is about 8x faster than JSDOM.

That said, it's also got a pretty significant npm footprint 

```
------------------------------------------------------
| Dependency            | Time     | Size   | # Deps |
------------------------------------------------------
| gulp-cheerio          | 40.5s    | 9.6 MB | 170    |
```

nathanmarks: IMO `JSDOM` is a better choice for testing code that interfaces with the browser as it exposes a browser-like API, not just a jQuery API subset for interfacing with a lightweight DOM.

For example, you can't do this with `cheerio`:

```js
import { jsdom } from 'jsdom';

const domDocument = jsdom('');
const myElement = domDocument.createElement('div');
myElement.setAttribute('data-my-element', '');

domDocument.body.appendChild(myElement);

const foundElement = domDocument.querySelector('[data-my-element]');

// true
console.log(foundElement === myElement);
```

It even says in the `cheerio` README

> Cheerio will not solve all your problems. I would still use JSDOM if I needed to work in a browser-like environment on the server, particularly if I wanted to automate functional tests.

AWaselnuk:
Anything I can do to help push this out the door? Can we save any of the remaining discussions for a future release?

tessalt:
I'd be cool with us merging this for now and adding more about functional/integration testing later.

testing/README.md
arapehl: My preference is `it()` / `describe()`. Especially since `it()` lends itself to much more fluid readability. I also like to steer clear of `it("should…"` in favour of being more direct like `it("returns a string", () => {`

nathanmarks:
@AWaselnuk There's a [small discussion](https://github.com/Shopify/javascript/pull/206#discussion_r76298253) RE cheerio vs jsdom that might be worth making a decision on.

AWaselnuk:
@nathanmarks People could benefit from most of this guide today. It is still helpful without the final decision on cheerio. Can we not open an issue or follow up PR for that debate to resolve it later? This PR brings us from 0 to 1 so anything we can do to ship sooner would be very good IMO.

testing/README.md
lemonmade: I'm fine with recommending JSDom 👍

lemonmade:
Two quick updates:

- Suggested using JSDom, I agree that it's probably the best recommendation

- After chatting with @minasmart and @mikkoh last week, I reorganized the guide and added some additional content that emphasized the features of testing tools we look for, rather than strictly suggesting a set of tools. The recommendations are still there for tools that someone who doesn't know/ doesn't care about JS testing, but for people who know what they are doing, have reasonable reasons not to use mocha/ chai/ sinon, and have a tool in mind that satisfies the recommended features/ testing style, I think it's fine for people to experiment a little.

cc/ @Shopify/javascript unless there's any objections I will get this initial version merged in.

AWaselnuk:
:ship: 😄

mikkoh:
👏 👏 👏

tessalt:
🎉 🎉 🎉

minasmart:
🎈 🎈 🎈

PR #208, opened 2016-08-08
Minor React styleguide updates

lemonmade:
This PR updates a few things on the React styleguide:

- Added back a warning for binding in `render`. Given all the performance talk lately, it's nice to encourage not binding so that we don't create extra rendering and we give components the ability to optimize using shallow comparisons of props.

- Recommended importing the things you need from React (`Component`, `PropTypes`, etc), instead of accessing them as properties on `React`.

- Updated recommendation for how to manage components in a folder with multiple subcomponents — do `Component/Component.js`, `Component/Subcomponent.js`, and export using `Component/index.js`, so that you get good file names everywhere and can still get easy imports elsewhere.

cc/ @Shopify/fed @bouk @harismahmood89

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -12,6 +12,7 @@ This guide provides a few guidelines on writing sensible React. Many of these ru
 1. [Components](#components)
 1. [Props](#props)
 1. [Testing](#testing)
+1. [Modules](#modules)
 1. [Resources](#resources)
 
 
@@ -26,11 +27,11 @@ This guide provides a few guidelines on writing sensible React. Many of these ru
 
   ```js
   // bad
-  class financialReport extends React.Component {}
+  class financialReport extends Component {}
   let MyReport = <financialReport />;
 
   // good
-  class FinancialReport extends React.Component {}
+  class FinancialReport extends Component {}
   let myReport = <FinancialReport />;
   ```
 
@@ -38,34 +39,14 @@ This guide provides a few guidelines on writing sensible React. Many of these ru
 
   ```js
   // bad
-  class BadComponent extends React.Component {
-    static propTypes = {my_prop: React.PropTypes.bool}
+  class BadComponent extends Component {
+    static propTypes = {my_prop: PropTypes.bool}
   }
 
   // good
-  class GoodComponent extends React.Component {
-    static propTypes = {myProp: React.PropTypes.bool}
+  class GoodComponent extends Component {
+    static propTypes = {myProp: PropTypes.bool}
   }
-  ```
-
-- [1.3](#1.3) <a name="1.3"></a> Try to define only one React component per file. Name that file the same as the component it exports, unless your component has static components that it also exports. In those cases, use `index.js` as the file that contains the main component for the directory.
-
-  ```js
-  // in Button.js
-  export default class Button extends React.Component {}
-
-  // in Field/Label.js
-  export default class Label extends React.Component {}
-
-  // in Field/index.js
-  import Label from './Label';
-
-  export default class Field extends React.Component {}
-  export {Label};
-
-  // in some other file...
-  import Button from './Button';
-  import Field, {Label} from './Field';
   ```
 
 [↑ scrollTo('#table-of-contents')](#table-of-contents)
@@ -246,7 +227,7 @@ This guide provides a few guidelines on writing sensible React. Many of these ru
 
   // good
   // (you can also just extend `Component` if you explicitly imported it)
-  class GoodComponent extends React.Component {
+  class GoodComponent extends Component {
     render() {
       return <div />;
     }
@@ -257,14 +238,14 @@ This guide provides a few guidelines on writing sensible React. Many of these ru
 
   ```js
   // bad
-  class BadComponent extends React.Component {
+  class BadComponent extends Component {
     render() {
       return React.createElement('div', {foo: 'bar'});
     }
   }
 
   // good
-  class GoodComponent extends React.Component {
+  class GoodComponent extends Component {
     render() {
       return <div foo="bar" />;
     }
@@ -273,7 +254,7 @@ This guide provides a few guidelines on writing sensible React. Many of these ru
 
 - [4.3](#4.3) <a name="4.3"></a> Prefer React’s stateless components for components that do not have any state or lifecycle hooks.
 
-  > **Note:** as of this writing, stateless components do not play nicely with React’s testing utilities. Until this is resolved, continue to extend `React.Component` for new components.
+  > **Note:** as of this writing, stateless components do not play nicely with React’s testing utilities. Until this is resolved, continue to extend `Component` for new components.
 
   > Why? Such components are very easy to represent as pure functions, and doing so encourages more stateless components, which decreases application complexity.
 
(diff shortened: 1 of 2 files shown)


bouk:
Can we recommend people to just drop prop types and use flow instead? It's so, so much nicer

react/README.md
```diff
@@ -498,22 +450,76 @@ This guide provides a few guidelines on writing sensible React. Many of these ru
...
+  }
+
+  // also good
+  class GoodComponent extends Component {
+    handleClick = this.handleClick.bind(this);
+    handleMouseEnter = this.handleMouseEnter.bind(this);
+
+    render() {
+      return <button onClick={this.handleClick} onMouseEnter={this.handleMouseEnter} />;
+    }
```
perrupa: did you mean to leave the `this.` in this last example? or did you mean:
```js
return <button onClick={handleClick} onMouseEnter={handleMouseEnter} />;

```

lemonmade: @perrupa nope, the properties above this are class instance properties, so they are still accessed with `this` (they just give you a quicker way of setting up instance properties that are constant expressions than to need to do it in the constructor).

perrupa: I guess I was expecting colons for assignment... My bad, carry on 😄

react/README.md
```diff
@@ -26,48 +27,28 @@ This guide provides a few guidelines on writing sensible React. Many of these ru
 
   ```js
   // bad
-  class financialReport extends React.Component {}
+  class financialReport extends Component {}
```
bouk: why this change? The react guide does `React.Component`, I think we should just follow that by default https://facebook.github.io/react/docs/reusable-components.html#es6-classes

react/README.md
```diff
@@ -498,22 +450,76 @@ This guide provides a few guidelines on writing sensible React. Many of these ru
...
+
+    render() {
+      return <button onClick={this.handleClick} onMouseEnter={this.handleMouseEnter} />;
+    }
+  }
+
+  // also good
+  class GoodComponent extends Component {
+    handleClick = this.handleClick.bind(this);
+    handleMouseEnter = this.handleMouseEnter.bind(this);
```
bouk: how about the following alternative:


```
handleMouseEnter = () => {
  ...
}
```

I'm quite a fan of it

react/README.md
```diff
@@ -528,6 +534,51 @@ This guide provides a few guidelines on writing sensible React. Many of these ru
...
 
+## Modules
+
+- [7.1](#7.1) <a name="7.1"></a> When importing from React, explicitly import the exact features you need (in addition to React, which will be a signal to the compiler to compile JSX).
+
+  > Why? Named imports explicitly indicate what you will be using at the top of the file, and reduce the duplication of accessing properties on `React`.
+
+  ```js
+  // bad
+  import React from 'react';
```
bouk: this is how every react guide does it, I think it's fine

react/README.md
```diff
@@ -528,6 +534,51 @@ This guide provides a few guidelines on writing sensible React. Many of these ru
...
+
+  ```js
+  // in Card/Section.js
+  export default class Section extends Component {}
+
+  // in Card/Card.js
+  import Section from './Section';
+
+  export default class Card extends Component {
+    static Section = Section;
```
bouk: I don't like this very much. I'd prefer renaming `Section` to `CardSection` so it's clear that it's related

react/README.md
```diff
@@ -528,6 +534,51 @@ This guide provides a few guidelines on writing sensible React. Many of these ru
...
+  }
+
+  // in Card/index.js
+  import Card from './Card';
```
bouk: why not just name it `Card/index.js`? Pretty clear what it means.

react/README.md
lemonmade: I saw that in your branch and I don't like the pattern — you end up having tons of components that are actually only ever supposed to be used as a child of a particular component but that isn't really explicit from the naming. I think `<Card.Section />` more clearly shows intent than `<CardSection />`, and is cleaner when there are lots of children components you might need.

react/README.md
lemonmade: Because then all of the files you are editing are named `index.js` and it's harder to work in the project, I've been using this setup and it makes it much easier to write things (also keeps it consistent with components that are just <Component>.js and don't need their own directory.

react/README.md
lemonmade: :-1:, when React inevitably does an ES6 version we will be happy that we are only importing what we need so that we can tree shake.

react/README.md
lemonmade: I'm kind of torn on this, I like the consistency of the methods being all declared in the same way but this pattern is kind of annoyingly boilerplate-ey. Only real thing I can see is that doing it in the way you suggest creates a (probably?) larger reference that needs to be stored per instance, probably not a big deal though.

react/README.md
bouk: all right I reverse my decision actually, the proper ES way of importing under namespace would be `import * as React from 'react'`, so what they're using in the guide is kind of outdated. nvm 👍

react/README.md
bouk: If we're going to bring up tree shaking, this also makes it impossible to tree shake out any components that are assigned to the static of another like this

react/README.md
bouk: not sure how the files being named index.js makes it harder, it's clear from the directory name what the file contains. It just adds clutter with an extra file that doesn't do much

lemonmade: I don't know what everyone's setup looks like, but in Sublime and Atom, files with the same name end up getting expanded out to show the full root-relative directory, which means you can maybe get a couple of `index.js` files open at a time before it becomes unreadable.

react/README.md
lemonmade: The reason I think this is different is because when components are connected (like Card and CardSection), they are effectively guaranteed to be used together, so you won't be able to eliminate much if anything anyways.

react/README.md
bouk: dunno if I did something funky with my sublime set up, but it only show the parent directory name next to it for me

react/README.md
bouk: I just don't really like it because it kind of goes around the ES6 module system and depends on adding properties to things that, although they're related, don't _need_ that property. It feels like kind of a step backwards

lemonmade: Would it make sense to do them as named exports, so you would have the default be the "root" component and then named for each subcomponent? Personally I still like the static properties because they make it super clear in JSX that it's a child, but maybe named exports is a reasonable compromise?

bouk: yeah, I think it makes more sense as named exports

react/README.md
lemonmade: @bouk how about we leave either one open (binding existing methods, or writing the methods as instance properties)?

react/README.md
lemonmade: (discussed below)

react/README.md
bouk: yeah I'm fine with both

lemonmade:
@bouk updated to recommend Flow (allowing proptypes since some projects use them, and you need to use them for setting context types IIRC), and to encourage components that expose multiple subcomponents to do them through named exports. Thoughts?

bouk:
:shipit:

PR #211, opened 2016-08-17
Linting updates

lemonmade:
This PR makes a few small updates to our linting config:

- Updated the `indent` rule to force `case` statements and member expressions nested on multiple lines to be indented one level (2 spaces). Currently, `case` is not supposed to be indented, and we have no lint rule that dictates multiline member expressions.
- Updated the `type-id-match` rule for flow to require all flow types to match `<Something>(Type|Props|State|Context)`, which I found to be a good setting to use for React stuff.
- Updated all the dependencies and configured the new rules (most are just turned off). The biggest addition is a rule that prevents using interpolation-like sequences in strings that aren't template literals (i.e., no `'${foo}'`, since this is probably an error).

cc/ @Shopify/javascript


packages/eslint-plugin-shopify/lib/config/rules/stylistic-issues.js
```diff
@@ -34,7 +36,7 @@ module.exports = {
   // Require identifiers to match the provided regular expression
   'id-match': 'off',
   // This option sets a specific tab width for your code
-  'indent': ['warn', 2],
+  'indent': ['warn', 2, {SwitchCase: 1, MemberExpression: 1}],
```
tessalt: 😍

bouk:
I disagree with indenting case statements. Unindented is the standard style in every other language, so I don't see why we should go against that

harismahmood89:
```
switch (action.type) {
  case types.UPDATE_TEXT: {
    const componentText = Object.keys(action.text).reduce((components, componentKey) => {
      components[componentKey] = _.merge({}, state[componentKey], action.text[componentKey]);
      return components;
    }, {});

    return Object.assign({}, state, componentText);
  }
  default: {
    return state;
  }
  }
```
you have to admit that this looks very strange. Text editors when set to JS also indent case statements by default, so there's some differences set for JS vs other languages

packages/eslint-plugin-shopify/lib/config/rules/stylistic-issues.js
harismahmood89: 👍

bouk:
It looks weird because you're putting brackets around case statements, which you don't need to (I've never seen that done before?)

harismahmood89:
We enforce the brackets to ensure lexical declarations only apply to a particular case: http://eslint.org/docs/rules/no-case-declarations

bouk:
whatever, I don't care enough about the case thing, :shipit:

PR #214, opened 2016-09-06
Always commas for multiline function params

bouk:
@lemonmade for consistency with the object dangle rules


lemonmade:
:+1: this has been bugging me too. We have the necessary transform in the babel preset to remove these, right?

bouk:
It seems the comma dangle transform is in the es2017 preset in babel, so we might need to update ours https://github.com/babel/babel/blob/master/packages/babel-preset-es2017/index.js

bouk:
Ah but it's also in the stage 3 preset, which we include in our shopify preset, so we're good

PR #215, opened 2016-09-07
Turn off `no-duplicate-imports` rule

bouk:
`import/no-duplicates` is better, as it handles Flow and different paths pointing to the same file.

@lemonmade

[More info about import/no-duplicates](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-duplicates.md)

### DIFF
--- a/packages/eslint-plugin-shopify/lib/config/rules/ecmascript-6.js
+++ b/packages/eslint-plugin-shopify/lib/config/rules/ecmascript-6.js
@@ -19,8 +19,8 @@ module.exports = {
   'no-const-assign': 'error',
   // Disallow duplicate name in class members
   'no-dupe-class-members': 'error',
-  // Disallow duplicate module imports
-  'no-duplicate-imports': 'error',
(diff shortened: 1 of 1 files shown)


packages/eslint-plugin-shopify/lib/config/rules/ecmascript-6.js
```diff
@@ -19,8 +19,8 @@ module.exports = {
   'no-const-assign': 'error',
   // Disallow duplicate name in class members
   'no-dupe-class-members': 'error',
-  // Disallow duplicate module imports
-  'no-duplicate-imports': 'error',
+  // Disallow duplicate module imports (disabled, as import/no-duplicates does the same job but better)
+  'no-duplicate-imports': 'off',
```
lemonmade: ❤️ I just ran into this and figured I was going to have to submit a PR for that rule. Awesome that we already had a better rule in place!

lemonmade:
🚀

PR #220, opened 2016-09-14
Update babel presets

lemonmade:
This PR updates how Shopify's Babel presets work. I know I kind of did this before and we decided not to separate them out so much, but I have some use cases where this makes sense and I feel like it is more in line with our linting configs.

Basically, it works like this:

- There are separate configs for Node, Web, and React projects, where React is totally usable with either of the other two. The default config is just the web config.

- The web config takes options, which it passes on to the `es2015` preset. This allows us to use the `{modules: false}` option for our config, which makes the preset work with Rollup and Webpack 2 tree shaking.

- The node option also takes options: `modules` (just like the web one), and `version` for specifying a desired version of Node. I was using a preset that just used `process.version` to check what plugins to add, but that didn't make sense since you usually develop in a more modern version of node than the one you build for. I just copied the logic into our own config so that you can manually specify the version.

I like this structure much better personally. It also enables things I've been working on: using Webpack 2 to do tree-shaked builds of both server and client bundles, both of which need the React transforms as well. The config just requires two lines instead of one for the client bundle, and the preset can actually be used for node without over-transforming stuff that doesn't need to be transformed:

```js
// client
{
  presets: [
    ['shopify/web', {modules: false}],
    'shopify/react',
  ],
};

// server
{
  presets: [
    ['shopify/node', {modules: false}],
    'shopify/react',
  ],
};
``` 

cc/ @Shopify/javascript @bouk thoughts?

### DIFF
--- a/packages/babel-preset-shopify/node.js
+++ b/packages/babel-preset-shopify/node.js
@@ -2,7 +2,7 @@ var semver = require('semver');
 
 module.exports = function shopifyNodePreset(context, options) {
   options = options || {};
-  var version = options.version || '5.7.0';
+  var version = options.version || process.version;
   var modules = options.modules == null ? true : options.modules;
   var pluginsList = [];
 
--- a/packages/babel-preset-shopify/node.js
+++ b/packages/babel-preset-shopify/node.js
@@ -1,6 +1,40 @@
-module.exports = {
-  presets: [
-    require('babel-preset-es2015-node'),
-    require('./non-standard-features'),
-  ],
+var semver = require('semver');
+
+module.exports = function shopifyNodePreset(context, options) {
+  options = options || {};
+  var version = options.version || '5.7.0';
+  var modules = options.modules == null ? true : options.modules;
+  var pluginsList = [];
+
+  if (modules) {
+    pluginsList.push(
+      require('babel-plugin-transform-es2015-modules-commonjs')
+    );
+  }
+
+  if (semver.lt(version, '6.0.0')) {
+    pluginsList.push(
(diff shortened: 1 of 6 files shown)


packages/babel-preset-shopify/node.js
```diff
@@ -1,6 +1,40 @@
-module.exports = {
-  presets: [
-    require('babel-preset-es2015-node'),
-    require('./non-standard-features'),
-  ],
+var semver = require('semver');
+
+module.exports = function shopifyNodePreset(context, options) {
+  options = options || {};
+  var version = options.version || '5.7.0';
```
lemonmade: Note to self: probably just default to `process.version`.

packages/babel-preset-shopify/non-standard-features.js
```diff
@@ -3,7 +3,6 @@ module.exports = {
     require('babel-preset-stage-2'),
   ],
   plugins: [
-    require('babel-plugin-transform-class-properties'),
```
lemonmade: Part of `stage-2`.

lemonmade:
Reping @Shopify/javascript — how do people feel about arranging the Babel presets in this way? Will this make our Babel preset usable by all projects you folks are currently working on? What can we do to make that happen, if the answer is no?

packages/babel-preset-shopify/node.js
bouk: yep, please do

bouk:
This makes sense with it. I'm OK with it as long as it's backwards-compatible

nathanmarks:
LGTM from a marketing/growth perspective.

PR #229, opened 2016-10-17
Add codemod for "flattening" iife global assignments

nathanmarks:
Let me know if you have a better idea for a name... 

This was created for preparing files in `marketing_assets` for some of the other transforms we have such as `global-assign-to-default-export`, etc...

It takes in your `appGlobalIdentifiers` (such as `ShopifyMarketing`) and lifts the code from inside the IIFE to the module root. It looks at the `return` statement inside the IIFE and assigns that + members of the returned object to the global identifier.

If the naming of the return identifier is inconsistent, it uses the property identifier found on the global object rather than the return identifier used inside the IIFE.

Every module in `marketing_assets` and other marketing properties are written using this pattern: https://github.com/Shopify/marketing_assets/blob/master/app/assets/javascripts/marketing_assets/modules/accordion.js

### DIFF
--- a/packages/shopify-codemod/test/fixtures/flatten-iife-global-assignment/inconsistent-naming.input.js
+++ b/packages/shopify-codemod/test/fixtures/flatten-iife-global-assignment/inconsistent-naming.input.js
@@ -1,0 +1,9 @@
+App.A = (function() {
+  function Q() {}
+
+  Q.prototype.foo = function() {
+    return 'bar';
+  };
+
+  return Q;
+})();
--- a/packages/shopify-codemod/test/fixtures/flatten-iife-global-assignment/inconsistent-naming.input.js
+++ b/packages/shopify-codemod/test/fixtures/flatten-iife-global-assignment/inconsistent-naming.input.js
@@ -5,5 +5,9 @@ App.A = (function() {
     return 'bar';
   };
 
+  function buzz() {
+    return new Q();
+  }
+
   return Q;
 })();
--- a/packages/shopify-codemod/transforms/flatten-iife-global-assignment.js
+++ b/packages/shopify-codemod/transforms/flatten-iife-global-assignment.js
@@ -1,4 +1,8 @@
-import {findFirstMember, getBlockStatementFromFunction} from './utils';
+import {
+  findFirstMember,
+  getBlockStatementFromFunction,
+  matchLast,
+} from './utils';
 
 export default function flattenIifeGlobalAssignment(
   {source},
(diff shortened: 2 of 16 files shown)


packages/shopify-codemod/transforms/flatten-iife-global-assignment.js
```diff
@@ -0,0 +1,113 @@
...
+      ),
+    );
+  }
+
+  function formatReturnAssignmentMemberExpression(memberExpression, prependee) {
+    const identifiers = [];
+    let currentNode = memberExpression;
+
+    // Take all but the last, which is copied from the base memberExpression
+    while (currentNode.type === 'MemberExpression') {
```
GoodForOneFare: There's a `removeFirstMember` function in the `global-assignment-to-default-export` transforms.  That may be usable/adaptable here.

GoodForOneFare:
Good stuff.  Also needs:
- Docs for the transform
- Changelog entry
- Added to the [esify transform list](https://github.com/Shopify/javascript/blob/master/packages/esify/lib/codemod.js)

packages/shopify-codemod/transforms/flatten-iife-global-assignment.js
```diff
@@ -0,0 +1,113 @@
...
+    return newExpression;
+  }
+
+  function flattenAssignment(path) {
+    const {left: member, right: statement} = path.node.expression;
+    const functionBlock = getBlockStatementFromFunction(statement.callee);
+
+    let returnIdentifier;
+
+    function liftNode(result, node) {
```
nathanmarks: Nope. I cheated and made this an anon function now.

packages/shopify-codemod/transforms/flatten-iife-global-assignment.js
```diff
@@ -0,0 +1,113 @@
...
+      .find(j.ExpressionStatement, {
+        expression: {
+          type: 'AssignmentExpression',
+          operator: '=',
+          left: {
+            type: 'MemberExpression',
+            object: (object) =>
+              appGlobalIdentifiers.indexOf(findFirstMember(object).name) >= 0,
+          },
+          right: {
```
nathanmarks: Thanks, awesome suggestion.

packages/shopify-codemod/test/fixtures/flatten-iife-global-assignment/inconsistent-naming.input.js
```diff
@@ -0,0 +1,9 @@
+App.A = (function() {
+  function Q() {}
+
+  Q.prototype.foo = function() {
+    return 'bar';
+  };
+
```
nathanmarks: Now that you mention it, got another couple of test cases that I want to handle just in case, for eg:

```
App.A = (function() {
  function A() {}

  A.prototype.woof = function() {}

  A.factoryFunction = function() {
    return new A();
  }

  return A.factoryFunction;
})();
```

GoodForOneFare:
Reopened this PR in the standalone `shopify-codemod` repo: https://github.com/Shopify/shopify-codemod/pull/1.  Closing here.

PR #233, opened 2016-11-17
Fix esify eslint

### LINKED ISSUE #185: ESLint is not actually running
My fix to ESLint broke it. Trying to do the right thing locally results in some crazy require-related errors, which I think relate to the babel stuff we do in `shopify-codemod`.

TylerHorth:
Attempt to load local eslint and process with --fix. 

Fixes #185 

@lemonmade @GoodForOneFare

### DIFF
--- a/packages/esify/lib/codemod.js
+++ b/packages/esify/lib/codemod.js
@@ -1,6 +1,6 @@
 var jscodeshift = require('jscodeshift');
 
-require('babel-register')({ignore: false});
+require('babel-register')({only: /node_modules\/(shopify-codemod|js-codemod)/});
 
 var TRANSFORMS = [
   {path: 'shopify-codemod/transforms/split-if-assignments'},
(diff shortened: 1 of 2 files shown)


lemonmade:
Code looks good 👍 

Can you check what the compile times before/ after are for a large-ish set of files (maybe try 10–20)? I want to make sure we don't regress too much on that measure.

GoodForOneFare:
🎩 went well, code's good.  🚢 assuming there isn't a significant perf regression.

TylerHorth:
I ran it on all coffeescript files in `assets/javascripts/admin/modules/products/` (~30 files), there wasn't a significant regression:

|eslint|real|user|sys|
|-----|----|----|----|
|NO|34.53s|30.72s|22.85s|
|YES|38.34s|34.60s|23.01s|

@lemonmade @GoodForOneFare 🚢 ?

GoodForOneFare:
So, like, ~125ms/file?  For the use case of many people updating one file at a time, looks worth it to me.

lemonmade:
🐑

PR #236, opened 2016-11-17
Bump eslint version

TylerHorth:
I don't believe there are any breaking changes. I tested esify quickly, it seems to work fine, and fixes more errors 👍 

@lemonmade @GoodForOneFare

### DIFF
--- a/packages/chai-jscodeshift/index.js
+++ b/packages/chai-jscodeshift/index.js
@@ -17,7 +17,7 @@ var config = {
 module.exports = function chaiJSCodeShift(options) {
   var finalConfig = merge({}, config, options || {});
 
-  // eslint-disable-next-line no-shadow, func-style
+  // eslint-disable-next-line no-shadow, func-style, func-name-matching
   var chaiHelper = function chaiJSCodeShift(chaiOptions) {
     var Assertion = chaiOptions.Assertion;
     var assert = chaiOptions.assert;
--- a/packages/chai-jscodeshift/index.js
+++ b/packages/chai-jscodeshift/index.js
@@ -17,8 +17,7 @@ var config = {
 module.exports = function chaiJSCodeShift(options) {
   var finalConfig = merge({}, config, options || {});
 
-  // eslint-disable-next-line no-shadow, func-style, func-name-matching
-  var chaiHelper = function chaiJSCodeShift(chaiOptions) {
+  function chaiHelper(chaiOptions) {
     var Assertion = chaiOptions.Assertion;
     var assert = chaiOptions.assert;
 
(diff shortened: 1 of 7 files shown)


packages/shopify-codemod/transforms/global-assignment-to-default-export.js
```diff
@@ -12,7 +12,7 @@ export default function globalAssignmentToDefaultExport({source}, {jscodeshift:
   }
 
   function isDefaultExportMemberAssignment(defaultExportName, assignmentTargetString) {
-    return new RegExp(`^${defaultExportName}\..+`).test(assignmentTargetString);
+    return new RegExp(`^${defaultExportName}\\..+`).test(assignmentTargetString);
```
TylerHorth: The new eslint caught this 👌 
Need to escape the escape in order for the regex to match the literal period.

GoodForOneFare: This looks useful, but should probably be in a separate PR.  It's easier to update CHANGELOG if we can just scan through commit messages.

GoodForOneFare:
Awesome! 🎆

VERSION
```diff
@@ -1 +1 @@
-15.0.3
\ No newline at end of file
+15.1.0
```
lemonmade: Don't edit this file directly, it will be edited when we publish the new version

packages/eslint-plugin-shopify/lib/config/rules/stylistic-issues.js
```diff
@@ -21,6 +21,8 @@ module.exports = {
   'eol-last': 'warn',
   // Disallow space between function identifier and application
   'func-call-spacing': 'warn',
+  // Require function names to match the name of the variable or property to which they are assigned
+  'func-name-matching': 'warn',
```
TylerHorth: @lemonmade How do you feel about this rule? http://eslint.org/docs/rules/func-name-matching

lemonmade: I know we also have a rule that prevents shadowing, if you do

```js
const foo = function foo() {}
```

With the new rules does it warn that the function `foo` is shadowing? If not, then 👍

TylerHorth: So it doesn't warn about shadowing, but it does complain about `func-style` since we have it set to declaration only. 

I suppose it couldn't hurt to have this rule as well, there may be some places that it covers where `func-style` doesn't.

lemonmade: :+1:

packages/chai-jscodeshift/index.js
```diff
@@ -17,7 +17,7 @@ var config = {
 module.exports = function chaiJSCodeShift(options) {
   var finalConfig = merge({}, config, options || {});
 
-  // eslint-disable-next-line no-shadow, func-style
+  // eslint-disable-next-line no-shadow, func-style, func-name-matching
   var chaiHelper = function chaiJSCodeShift(chaiOptions) {
```
lemonmade: Rename the function here to `chaiHelper` and we should be able to ditch the `no-shadow` and `func-name-matching` eslint disables.

GoodForOneFare:
🚢

PR #237, opened 2016-11-18
Warn when removing unused expression

### LINKED ISSUE #225: esify should warn about removing unused statements
@beefchimi noticed that esify is removing `@node.offsetHeight` from this block of code:

``` coffeescript
      @node.addEventListener 'transitionend', @_finishTransition
      @node.classList.add('collapsible--has-transition')
      @node.style.height = @_getHeight()
      @node.offsetHeight # reflows the node to prevent transition jank
```

It's an unused statement, but it performs a useful function.  I think esify should -- at least -- warn people about this (and maybe we should come up with a convention for forcing a reflow).

TylerHorth:
Resolves #225 

screen shot 2016-11-18 at 1 52 29 pm

@lemonmade @GoodForOneFare


packages/shopify-codemod/transforms/utils.js
```diff
@@ -1,4 +1,63 @@
 import j from 'jscodeshift';
+import chalk from 'chalk';
+
+function range(file, node, source, color, CONTEXT = 2) {
```
lemonmade: What is `CONTEXT` here? Do we actually need it to be a parameter?

packages/shopify-codemod/transforms/utils.js
```diff
@@ -1,4 +1,63 @@
...
+  if (file === null || node === null || source === null || color == null) {
```
lemonmade: This checks for explicitly passing `null`. If you want to bail on either explicitly passing `null` or just omitting the arguments, you should use `== null`.

packages/shopify-codemod/transforms/utils.js
```diff
@@ -1,4 +1,63 @@
...
+    return '';
+  }
+
+  const lines = source.split(/\r\n|[\n\v\f\r\x85\u2028\u2029]/);
```
lemonmade: Can you move the regex outside the function and give it a meaningful variable name?

packages/shopify-codemod/transforms/utils.js
```diff
@@ -1,4 +1,63 @@
...
+
+  const sl = node.node.loc.start.line - 1;
```
lemonmade: Use destructuring to get at the pieces of this you care about:

```js
const {start, end} = node.node.loc;
const sl = start.line - 1;
// ...
```

packages/shopify-codemod/transforms/utils.js
```diff
@@ -1,4 +1,63 @@
...
+  const sc = node.node.loc.start.column;
```
lemonmade: Use full words (`startColumn` instead of `sc`).

packages/shopify-codemod/transforms/utils.js
lemonmade: This function name does not really describe what it actually does.

packages/shopify-codemod/transforms/utils.js
```diff
@@ -1,4 +1,63 @@
...
+        paintStart = 0;
+      }
+      if (i === el) {
+        paintEnd = ec;
+      } else {
+        paintEnd = Number.MAX_SAFE_INTEGER;
+      }
+    }
+
+    result += `    ${i + 1}`.slice(-4);
```
lemonmade: This right pads to four digits, right? Might be nice to have a little utility method for this, it's not a super common idiom and giving it a name will help the next person reading this.

packages/shopify-codemod/transforms/utils.js
```diff
@@ -1,4 +1,63 @@
...
+  const lines = source.split(/\r\n|[\n\v\f\r\x85\u2028\u2029]/);
+
+  const sl = node.node.loc.start.line - 1;
+  const sc = node.node.loc.start.column;
+  const el = node.node.loc.end.line - 1;
+  const ec = node.node.loc.end.column;
+  const start = Math.max(0, sl - CONTEXT);
+  const end = Math.min(lines.length, el + CONTEXT);
+
+  let result = `(${file})\n`;
```
lemonmade: I feel like this doesn't belong in this function given that this function is already doing a lot of work specifically to format the source code. Would also help reduce the number of params you need to pass in.

packages/shopify-codemod/transforms/utils.js
```diff
@@ -1,4 +1,63 @@
...
+
+  for (let i = start; i <= end; i++) {
```
lemonmade: You should be able to do this using some native array methods:

```js
return lines.map((line) => {
  // create the new string for this line
  return '';
}).join('\n');
```

packages/shopify-codemod/transforms/utils.js
```diff
@@ -1,4 +1,63 @@
...
+    result += `    ${i + 1}`.slice(-4);
+    result += line.slice(0, paintStart);
+    result += color(line.slice(paintStart, paintEnd));
+    result += line.slice(paintEnd, Number.MAX_SAFE_INTEGER);
+    result += '\n';
+  }
+  return result;
+}
+
+export function info(message, file = null, node = null, source = null) {
```
lemonmade: No need to `= null` for these. In JS, if an argument is omited, it is just set to `undefined`, which is fine here.

packages/shopify-codemod/transforms/utils.js
TylerHorth: `CONTEXT` is the number of lines before and after the node to display. I put it in a parameter just to avoid having a magic number inside the function, but I could also make it a constant if you think that would be better.

lemonmade: Either constant out of the function scope, or a regular parameter inside the function scope (if you are doing that, it would probably also be good to make these an object instead of positional arguments).

packages/shopify-codemod/transforms/utils.js
```diff
@@ -1,4 +1,73 @@
...
+  const {start, end} = node.node.loc;
+  const startLine = start.line - 1;
+  const startColumn = start.column;
+  const endLine = end.line - 1;
+  const endColumn = end.column;
+  const contextStart = Math.max(0, startLine - CONTEXT_LINES);
+  const contextEnd = Math.min(lines.length, endLine + CONTEXT_LINES);
+
+  let result = '';
+  for (let i = contextStart; i <= contextEnd; i++) {
```
TylerHorth: I didn't change this to map since we still need access to the index, and I think this way is a bit cleaner than: 
```js
return lines.slice(contextStart, contextEnd + 1).map((line, index) => {
  const index = index + contextStart;
  // do stuff
  return new_line;
}).join('\n');
```

packages/shopify-codemod/transforms/utils.js
```diff
@@ -1,4 +1,73 @@
 import j from 'jscodeshift';
+import chalk from 'chalk';
```
GoodForOneFare: This new code doesn't fit with what's already in the file (it's mostly AST element matchers/creators).  I'd prefer seeing it in a separate file (`code-utils.js`, or something).

packages/shopify-codemod/transforms/utils.js
GoodForOneFare: I'll add more detailed thoughts tonight, but my initial impression is that I'm relying on guesswork to know what's happening in here.  Splitting this into separate calculate/render functions might help.

Test would have been nice, but writing them after the fact doesn't feel useful now that this is working.

packages/shopify-codemod/transforms/utils.js
```diff
@@ -1,4 +1,73 @@
...
+}
+
+function logMessage(label, message, color, file, node, source) {
+  const coloredLabel = color(`[${label}]`);
+  const fileTag = file ? ` (${file.replace(/\.coffee$/, '.js')})` : '';
+  const context = node && source ? `\n${codeContext(node, source, color)}` : '';
+  return `${coloredLabel} ${message}${fileTag}${context}`;
+}
+
+export function info(message, file, node, source) {
```
GoodForOneFare: This isn't used, right?  I'd delete it for now.

packages/shopify-codemod/transforms/utils.js
```diff
@@ -1,4 +1,73 @@
...
+  // eslint-disable-next-line no-console
+  console.info(logMessage('info', message, chalk.blue, file, node, source));
+}
+
+export function warn(message, file, node, source) {
+  // eslint-disable-next-line no-console
+  console.warn(logMessage('warning', message, chalk.yellow, file, node, source));
+}
+
+export function error(message, file, node, source) {
```
GoodForOneFare: Delete this, too.

TylerHorth:
@GoodForOneFare I did some refactoring to make the code easier to read, and added an property whitelist. Could you take another look?

packages/shopify-codemod/transforms/console-utils.js
```diff
@@ -0,0 +1,68 @@
...
+
+function paintLine(line, from, to, color) {
+  return `${line.slice(0, from)}${color(line.slice(from, to))}${line.slice(to)}`;
+}
+
+function formatLine(line, lineNum) {
+  return `${chalk.gray(padLeft(lineNum + 1, 4))}|${line}\n`;
+}
+
+function codeContext(node, source, color) {
```
GoodForOneFare: This is better, but still does too much for me.  My brain is occupied with splitting the "figure out which lines to display" from the "display the lines" parts.

So figuring out the lines is probably better in its own function, like:
```js
function getContextLines(loc, source) {
  const lines = source.split(NEWLINE_REGEX);
  const {start, end} = loc;
  const startLineIndex = loc.start.line - 1;
  const endLineIndex = loc.end.line - 1;

  const contextStart = Math.max(0, startLineIndex - CONTEXT_LINES);
  const contextEnd = Math.min(lines.length, endLineIndex + CONTEXT_LINES);

  return lines.slice(contextStart, contextEnd).map((line, i) => {
    const lineIndex = i + contextStart;
    const isHighlighted = (lineIndex >= startLineIndex) && (lineIndex <= endLineIndex);

    return {
      line,
      lineIndex,
      isHighlighted,
      ...getHighlightRange(isHighlighted, start.column, end.column),
    };
  });
}


function getHighlightRange(isHighlighted, startColumn, endColumn) {
  const highlightStart = isHighlighted ? startColumn : Number.MAX_SAFE_INTEGER;
  const highlightEnd = isHighlighted ? endColumn : Number.MAX_SAFE_INTEGER;

  return {highlightStart, highlightEnd};
}
```

Note that the result contains more than just strings.  That allows us to simplify display a little:

```js
function codeContext(node, source, color) {
  const lines = getContextLines(node.get('loc').value, source);

  return lines.map((lineInfo) =>
    formatLine(
      paintLine(lineInfo, color),
      lineInfo.lineIndex + 1,
    ),
  ).join('');
}

function paintLine({line, highlightStart, highlightEnd}, color) {
  const prefix = line.slice(0, highlightStart);
  const highlighted = line.slice(highlightStart, highlightEnd);
  const suffix = line.slice(highlightEnd);

  return `${prefix}${color(highlighted)}${suffix}`;
}
```

GoodForOneFare: You can even split the "get context lines" and "add highlight info" into separate functions, but it gets a bit verbose.

PR #241, opened 2016-11-22
split-if-assignments should handle `else if` assignments

GoodForOneFare:
Previously, `else if (foo = bar)` would throw an exception because the transform tried to move the assignment to before the `else`.  That would've created an invalid AST.

This updates the logic to rewrite `else if (foo = bar) {}` to 
```
else { 
  foo = bar; 
  if (foo) {

  }
}
```

Please review @lemonmade, @TylerHorth


packages/shopify-codemod/test/fixtures/split-if-assignments/if-else-if.input.js
```diff
@@ -0,0 +1,11 @@
...
+  foo();
+} else if (b = 2) {
+  bar();
+} else if (c = 3) {
+  baz();
+} else if (d = 4) {
+  baz();
+} else {
+  qux();
+}
```
lemonmade: Can you add a test with nested `if else`?

```js
if (a = 1) {
  if (b = 2) {

  } else if (c = 3) {

  }
} else if (d = 4) {

}
```

packages/shopify-codemod/transforms/split-if-assignments.js
```diff
@@ -1,13 +1,32 @@
...
-      const assignment = path.get('test');
-      path.insertBefore(j.expressionStatement(assignment.node));
-      assignment.replace(assignment.node.left);
-    })
-    .toSource(printOptions);
+
+  const sourceAST = j(source);
+
+  // Handle else-ifs in reverse order (otherwise initial transforms orphan later paths).
+  sourceAST.find(j.IfStatement, {
```
lemonmade: `.find` on its own line, aligned with `.paths()`

packages/shopify-codemod/transforms/split-if-assignments.js
```diff
@@ -1,13 +1,32 @@
...
+    alternate: {
+      type: j.IfStatement.name,
+      test: j.AssignmentExpression.check,
+    },
+  }).paths()
+    .reverse()
+    .forEach(({node}) => {
+      const elseIf = node.alternate;
+      const test = elseIf.test;
+      node.alternate = j.blockStatement([
```
lemonmade: I am a little rusty on this, but I seem to remember preferring using the path and using `replace()` instead of directly mutating the node.

packages/shopify-codemod/transforms/split-if-assignments.js
```diff
@@ -1,13 +1,32 @@
...
+        j.expressionStatement(test),
+        j.ifStatement(test.left, elseIf.consequent, elseIf.alternate),
+      ]);
+    });
+
+  // Hoist assignments out of the first `if` and into the block body.
+  return sourceAST.find(j.IfStatement, {
```
lemonmade: Same as above

packages/shopify-codemod/test/fixtures/split-if-assignments/if-else-if.output.js
```diff
@@ -0,0 +1,24 @@
...
+if (a) {
+  foo();
+} else {
+  b = 2;
+
+  if (b) {
+    bar();
+  } else {
+    c = 3;
+
```
TylerHorth: Why are these newlines being added? Is this intentional?

GoodForOneFare: `recast` is adding them.  I don't think there's much that we can do about that.

GoodForOneFare:
@lemonmade please take another look.  I've cleaned up indentation, and used `path.replace` instead of direct node assignment.

PR #242, opened 2016-11-23
Split up sequence assignments generated by decaf

GoodForOneFare:
`decaf` takes stuff like `{@foo, @bar} = opts` and turns it into a sequence like `foo = opts.food, bar = opts.bar, opts`.  This PR:
* Converts each assignment to its own expression
* Removes the trailing `opts`

Note, I haven't handled some edge cases because they have no value to Shopify's current codebase:
* Transforming `return foo = opts.foo, opts` to `foo = opts.foo; return opts`
* Checking that all assigned fields are identifiers with a name matching their source property

Please review @lemonmade, @TylerHorth

### DIFF
--- a/packages/shopify-codemod/docs/split-assignment-sequences.md
+++ b/packages/shopify-codemod/docs/split-assignment-sequences.md
@@ -11,7 +11,7 @@ jscodeshift -t shopify-codemod/transforms/split-assignment-sequences <file>
 ```js
 foo = options.foo, bar = options.bar, options;
 
-this.foo = opts.foo, bar = opts.bar, options;
+this.foo = opts.foo, bar = opts.bar, opts;
 
 // BECOMES:
 
--- a/packages/shopify-codemod/docs/split-assignment-sequences.md
+++ b/packages/shopify-codemod/docs/split-assignment-sequences.md
@@ -1,0 +1,23 @@
+### `split-assignment-sequences`
+
+Splits `decaf`-generated assignment sequences into individual expressions, and removes their useless trailing return value.
+
+```sh
+jscodeshift -t shopify-codemod/transforms/split-assignment-sequences <file>
+```
+
+#### Example
+
+```js
+foo = options.foo, bar = options.bar, options;
+
+this.foo = opts.foo, bar = opts.bar, options;
+
+// BECOMES:
+
+foo = options.foo
+bar = options.bar
+
+this.foo = opts.foo
+this.bar = opts.bar
+```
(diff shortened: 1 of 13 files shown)


packages/shopify-codemod/docs/split-assignment-sequences.md
```diff
@@ -0,0 +1,23 @@
...
+```sh
+jscodeshift -t shopify-codemod/transforms/split-assignment-sequences <file>
+```
+
+#### Example
+
+```js
+foo = options.foo, bar = options.bar, options;
+
+this.foo = opts.foo, bar = opts.bar, options;
```
lemonmade: Trailing `opts`?

packages/shopify-codemod/transforms/split-assignment-sequences.js
```diff
@@ -0,0 +1,29 @@
...
+      const lastExpression = expressions[expressions.length - 1];
+      const otherExpressions = expressions.slice(0, -1);
+
+      return j.Identifier.check(lastExpression) &&
+        otherExpressions.every(j.AssignmentExpression.check) &&
+        otherExpressions.every((expression) =>
+          expression.right.object.name === lastExpression.name,
+        );
+    })
+    .replaceWith((path) => {
```
lemonmade: `forEach` and call `replace(null)` at the end? It feels awkward to prepend in `replaceWith`.

TylerHorth:
Is it possible to translate this to object destructuring?

----
```js
foo = opts.foo, bar = opts.bar, opts
```
=>
```js
({foo, bar} = opts);
```
----
```js
this.foo = opts.foo, this.bar = opts.bar, opts
```
=>
```js
({foo: this.foo, bar: this.bar} = opts);
```

GoodForOneFare:
@lemonmade examples of Coffee->decaf->JS that generates these:

`card_consumption`:
```coffee
class Shopify.CardConsumption
  constructor: (options) ->
    { @requestId } = options
```

becomes:

```js
    this.requestId = options.requestId, options;
```

`notifications.coffee` contains `{@iconPath, @shopName} = opts`

`inventory_options` contains 
```coffee
   {
      @onAdjustmentUpdated,
      @onQuantityUpdated,
      @onAdjustmentIncremented,
      @onAdjustmentDecremented
    } = options
```

GoodForOneFare:
@TylerHorth that's possible, but I find the es5 output to be way more readable ¯\_(ツ)_/¯  e.g.,

```js
this.foo = opts.foo;
this.bar = opts.bar;
```

... has the same character count as the deconstructed version, but I guarantee you that anyone in #fed will understand the es5 version in half the time.

GoodForOneFare:
@lemonmade please take another look.  Example Coffee provided, docs fixed, replace/insert/null changed to forEach/insert/replace.

Issue #253, opened 2016-12-19
Add yarn installation instructions


lemonmade:
Treat it as the default, but include npm instructions.

PR #254, opened 2016-12-19
Added guides for using Yarn to install deps

AWaselnuk:
Closes https://github.com/Shopify/javascript/issues/253

This PR adds instructions for using Yarn to install some deps.


AWaselnuk:
@lemonmade team work makes the dream work!

PR #257, opened 2017-01-17
Remove packages with solo repos

GoodForOneFare:
First commit removes packages, second commit removes monorepo-related files.

New repos:
https://github.com/Shopify/babel-preset-shopify
https://github.com/Shopify/chai-jscodeshift
https://github.com/Shopify/eslint-plugin-shopify
https://github.com/Shopify/shopify-codemod


lemonmade:
:+1: Going to need to remove the circle integration though.

GoodForOneFare:
It looks like deleting a Circle project isn't possible: https://discuss.circleci.com/t/add-ability-to-delete-a-project-from-circleci/936/6 😕

For now, I've just added a dumb `echo "No tests"` command to the Circle config, but even that spins up an environment that we don't need.

PR #267, opened 2017-11-27
Update JS testing guide to include Jest and info on generating coverage

vividviolet:
Revising the guide to add more information on Karma, Jest and generating test coverage.

I think the recommendation for Teaspoon is out of date since there are a bunch of drawbacks compared to Karma and Jest. I've successfully worked with @macdonaldr93 and a bunch of other FEDS to move existing projects off of Teaspoon and onto Karma which enabled easier testing and code coverage to be generated.

### DIFF
--- a/testing/README.md
+++ b/testing/README.md
@@ -41,17 +41,17 @@ If you feel that mocha/ sinon/ chai do not do a good job of addressing the way y
 
 - [testdouble](https://github.com/testdouble/testdouble.js) is a great alternative to sinon as a test double library.
 
-If your project is using [ReactJS](https://reactjs.org/) or is built with [Sewing-kit](https://github.com/Shopify/sewing-kit), we recommend using the following tools:
-- [Jest](http://facebook.github.io/jest/) for the test framework. 
-- [Enzyme](http://airbnb.io/enzyme/) for handling spies and mocks.
+If your project is using [ReactJS](https://reactjs.org/) or is built with [sewing-kit](https://github.com/Shopify/sewing-kit), we recommend using the following tools:
+- [Jest](http://facebook.github.io/jest/) for the test framework, which also handles spies and mocks.
+- [Enzyme](http://airbnb.io/enzyme/) for handling the rendering and manipulating React components.
 
 ### Test Runners
 
 In addition to the testing framework, you generally need a test *runner*. Which runner to use will depend on the nature of your project:
 
-- For projects using [Jest](http://facebook.github.io/jest/), you do not need an additional test runner. Just use Jest to run your tests and generate test coverage. See [Coverage](#coverage) for more information on test coverage with Jest.  
+- For projects using [Jest](http://facebook.github.io/jest/), you do not need an additional test runner - just use Jest to run your tests and generate test coverage. See [Coverage](#coverage) for more information on test coverage with Jest. Jest will default to providing you a "fake" browser environment with [JSDom](https://github.com/tmpvar/jsdom). If you want to test things in a true browser environment it will not be sufficient to rely on Jest alone as this is not currenlty supported.
 
-- For existing projects that already have tests written in mocha/sinon/chai, we recommend [Karma](https://karma-runner.github.io/1.0/index.html). It allows running tests in the browser and integrates well with other tools/frameworks you are likely using, such as bundlers and code coverage tools. See [Coverage](#coverage) for more information on test coverage with Karma. Feel free to instead use [Testem](https://github.com/testem/testem), another great JavaScript test runner, if you would like to try something a little different.
+- For existing projects that already have tests written in mocha/sinon/chai, we recommend [Karma](https://karma-runner.github.io/1.0/index.html). It allows running tests in the browser and integrates well with other tools/frameworks you are likely using, such as bundlers and code coverage tools. See [Coverage](#coverage) for more information on test coverage with Karma.
 
 ### Other
 
@@ -409,7 +409,7 @@ Here are some recommended ways to generate coverage:
 - **Jest**: You can use the built in coverage reporter by adding the `--coverage` argument when running tests. See (documentation)[https://facebook.github.io/jest/docs/en/cli.html#coverage].
 - **Other**: We recommend using [nyc](https://github.com/istanbuljs/nyc) as a standalone command line tool to generate coverage.
 
-It is best practice to generate coverage for all relevant JS files as opposed to only tested files. This will files are not inadvertently missed. There are a couple of ways to generate coverage for all files:
+It is best practice to generate coverage for all relevant JS files as opposed to only tested files. This will ensure files are not inadvertently missed. There are a couple of ways to generate coverage for all files:
 - **Karma**: Add the [`includeAllSources`](https://github.com/karma-runner/karma-coverage/blob/master/docs/configuration.md#includeallsources) option to your Karma `coverageReporter` config:
 ```js
 files: [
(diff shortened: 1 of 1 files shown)


testing/README.md
```diff
@@ -40,21 +41,23 @@ If you feel that mocha/ sinon/ chai do not do a good job of addressing the way y
...
+- For projects using [Jest](http://facebook.github.io/jest/), you do not need an additional test runner. Just use Jest to run your tests and generate test coverage. See [Coverage](#coverage) for more information on test coverage with Jest.  
 
-- Rails projects should use [Teaspoon](https://github.com/modeset/teaspoon), which provides an excellent GUI for running tests in the browser. If you are also using [sprockets-commoner](https://github.com/Shopify/sprockets-commoner) to transpile your JavaScript, make sure to include [teaspoon-bundle](https://github.com/Shopify/teaspoon-bundle) to generate your test bundle.
+- For existing projects that already have tests written in mocha/sinon/chai, we recommend [Karma](https://karma-runner.github.io/1.0/index.html). It allows running tests in the browser and integrates well with other tools/frameworks you are likely using, such as bundlers and code coverage tools. See [Coverage](#coverage) for more information on test coverage with Karma. Feel free to instead use [Testem](https://github.com/testem/testem), another great JavaScript test runner, if you would like to try something a little different.
```
nwtn: what’s the advantage of keeping testem here? are we using it anywhere? i'd be happy to just remove that bit.

testing/README.md
```diff
@@ -390,9 +393,55 @@ Visual regression testing allows you to be confident that changes made to the co
...
+
+Generating a test coverage report can help you identify areas of the code that are lacking tests. Coverage reports can be integrated with CodeCov and Circle CI to ensure sufficient coverage with every checkin. See the [CodeCov Integration Guide](https://vault.shopify.com/Codecov-integration) for more information.
+
+Here are some recommended ways to generate coverage:
+
+- **Karma**: You can use the [Karma Istanbul Reporter plugin](https://github.com/mattlewis92/karma-coverage-istanbul-reporter) to automatically generate coverage when tests are running.
+- **Jest**: You can use the built in coverage reporter by adding the `--coverage` argument when running tests. See (documentation)[https://facebook.github.io/jest/docs/en/cli.html#coverage].
+- **Other**: We recommend using [nyc](https://github.com/istanbuljs/nyc) as a standalone command line tool to generate coverage.
+
+It is best practice to generate coverage for all relevant JS files as opposed to only tested files. This will files are not inadvertently missed. There are a couple of ways to generate coverage for all files:
```
nwtn: Word missing in second sentence

testing/README.md
```diff
@@ -390,9 +393,55 @@ Visual regression testing allows you to be confident that changes made to the co
...
-
 ## React
 
 If your project is using React, you can find React-specific testing recommendations in the [React styleguide](https://github.com/Shopify/javascript/tree/master/react#testing).
 
 [↑ scrollTo('#table-of-contents')](#table-of-contents)
+
+## Coverage
+
+Generating a test coverage report can help you identify areas of the code that are lacking tests. Coverage reports can be integrated with CodeCov and Circle CI to ensure sufficient coverage with every checkin. See the [CodeCov Integration Guide](https://vault.shopify.com/Codecov-integration) for more information.
```
nwtn: might be worth mentioning that we’re not striving for 100% coverage usually. as you say, generating coverage reports should help us find untested code, but usually we don’t set a hard number like “we must have 90% coverage”.

testing/README.md
```diff
@@ -40,21 +41,23 @@ If you feel that mocha/ sinon/ chai do not do a good job of addressing the way y
 
 - [testdouble](https://github.com/testdouble/testdouble.js) is a great alternative to sinon as a test double library.
 
+If your project is using [ReactJS](https://reactjs.org/) or is built with [Sewing-kit](https://github.com/Shopify/sewing-kit), we recommend using the following tools:
+- [Jest](http://facebook.github.io/jest/) for the test framework. 
+- [Enzyme](http://airbnb.io/enzyme/) for handling spies and mocks.
```
lemonmade: Enzyme doesn't handle spies or mocks, it is purely for giving you a nice API for dealing with React components rendered in test. Jest provides its own spying/ mocking utilities that we use.

testing/README.md
```diff
@@ -40,21 +41,23 @@ If you feel that mocha/ sinon/ chai do not do a good job of addressing the way y
...
+
 ### Test Runners
 
 In addition to the testing framework, you generally need a test *runner*. Which runner to use will depend on the nature of your project:
 
-- For pure-JavaScript projects that do not need access to the DOM, avoid using a test runner and simply run the CLI for your test framework instead (all of the projects listed above have some form of CLI). If you need a simple DOM without actual rendering, you can also use [JSDom](https://github.com/tmpvar/jsdom) to construct a fake DOM.
-
-- For Non-Rails projects that require a full DOM, we recommend [Karma](https://karma-runner.github.io/1.0/index.html). It integrates well with other tools you are likely using, such as bundlers and code coverage tools. Feel free to instead use [Testem](https://github.com/testem/testem), another great JavaScript test runner, if you would like to try something a little different.
+- For projects using [Jest](http://facebook.github.io/jest/), you do not need an additional test runner. Just use Jest to run your tests and generate test coverage. See [Coverage](#coverage) for more information on test coverage with Jest.
```
lemonmade: Good to call out here: Jest will default to providing you a "fake" browser environment with JSDom. This is enough in most cases, but if you want to test things in a true browser environment it will not be sufficient to rely on Jest alone (unsure if Jest can be run in the browser yet, it couldn't in the past).

testing/README.md
lemonmade: Me too, never really made too much sense to me that we recommended two things.

testing/README.md
lemonmade: Some teams do have a strict coverage threshold, I think it's fair for that to be decided on a team-by-team basis.

testing/README.md
```diff
@@ -390,9 +393,55 @@ Visual regression testing allows you to be confident that changes made to the co
...
+- **Karma**: Add the [`includeAllSources`](https://github.com/karma-runner/karma-coverage/blob/master/docs/configuration.md#includeallsources) option to your Karma `coverageReporter` config:
+```js
+files: [
+  'app/assets/javascripts/**/*.js', // Regexp matching relevant files
+],
+coverageReporter: { 
+  includeAllSources: true
+}
+```
+- **Jest**: If you are using [Sewing-kit](https://github.com/Shopify/sewing-kit), coverage for all files should be enabled automatically. Otherwise, you can configure the [`collectCoverageFrom`](http://facebook.github.io/jest/docs/en/configuration.html#collectcoveragefrom-array) option:
```
lemonmade: What is otherwise here? When is coverage not collected?

testing/README.md
```diff
@@ -390,9 +393,55 @@ Visual regression testing allows you to be confident that changes made to the co
...
+  "collectCoverageFrom": [
+    '**/*.{js,jsx}', // Glob pattern for matching/excluding files
+    "!**/node_modules/**",
+    "!**/vendor/**"
+  ]
+}
+```
+- **Other**: Alternatively, you can use a single test entry file that loads all relevant JS modules. This is not recommended since the entry file would have to be updated whenever a new module is added.
+
+If you are writing code in ES6 and are bundling your code, it is best to generate coverage on non-bundled and non-transpiled code for better understandability. This should already be supported by default if you are using Webpack + Babel to bundle your code. If you are using Browserify for bundling, you might need to use the `es2016` preset along with the [browserify-istanbul plugin](https://github.com/devongovett/browserify-istanbul) to generate coverage for non-bundled and non-transpiled code. For example:
```
lemonmade: I don't think we do(/should) use browserify on many projects.

testing/README.md
vividviolet: I'm happy to get rid of it

testing/README.md
vividviolet: How about: 
> Generating a test coverage report can help you identify untested code. Coverage reports can be integrated with CodeCov and Circle CI to ensure team-specific coverage targets are met with every checkin.

testing/README.md
vividviolet: Oops, that's a mistake. I'll clarify.

testing/README.md
vividviolet: I meant to say, if you're not using Sewing-kit, you might need to set the `collectCoverageFrom` option yourself.

vividviolet:
@lemonmade I've addressed all issues. Can you take another look? Hoping to merge this as @macdonaldr93 is looking to create a link to this JS coverage info on the Vault page for [CodeCov Integration](https://vault.shopify.com/Codecov-integration)

PR #269, opened 2018-08-09
Add legacy CoffeeScript guide

nwtn:
I’ve been asked to shut down https://github.com/Shopify/CoffeeScript-Style-Guide if it’s not being used. It’s not _really_ but I think we should preserve this content for anybody updating legacy CoffeeScript which still, unfortunately, exists.

Other than the warning at the top, all of this is directly ported from the CofeeScript-Style-Guide repo unchanged.

I know all this stuff is going to eventually get cleaned up and moved elsewhere, but for the time-being this felt like the best home. Lmk if you disagree.

### DIFF
--- a/coffeescript/README.md
+++ b/coffeescript/README.md
@@ -1,0 +1,722 @@
+# CoffeeScript Style Guide
+
+**Note: CoffeeScript is deprecated at Shopify, please do not write any new CoffeeScript files.** Only use this guide if you’re updating existing CoffeeScript and converting is out of scope.
+
+## Intro
+
+This guide presents a collection of best-practices and coding conventions for the [CoffeeScript](http://jashkenas.github.com/coffee-script/) programming language, and describes how we write CoffeeScript at Shopify.
+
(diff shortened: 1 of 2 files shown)


coffeescript/README.md
```diff
@@ -0,0 +1,723 @@
+# CoffeeScript Style Guide
+
+Don’t write new CoffeeScript please. Only use this guide if you’re updating existing CoffeeScript and converting is out of scope.
```
TheMallen: Can we bolt this or something? Or maybe get a small bolded blurb that says something like `Note: CoffeeScript is deprecated at Shopify, please do not write any new CoffeeScript files.`

GoodForOneFare:
The content is still useful, so I'd just leave the repo up ¯\\_(ツ)_/¯

nwtn:
> The content is still useful, so I'd just leave the repo up ¯\(ツ)/¯

We’re hitting the limits of the number of private repos we’re allowed on GH and this isn’t worth wasting one on imo, especially since we already have the javascript and web-foundations repos.

nwtn:
Right now it’s basically Yet Another Place for people to have to look to find our docs. I’d prefer doing consolidation if possible, rather than making this lone wolf public.

So, let me approach this another way. Does anybody strongly object to me merging this? If not, I’ll merge and remove the other repo at the end of the day.

PR #270, opened 2018-10-25
Update 10.2 as per func-style allowArrowFunction

sambostock:
This updates rule 10.2 following the change in https://github.com/Shopify/eslint-plugin-shopify/pull/188 allowing arrow function expressions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1225,6 +1225,9 @@ npm run lint
 
   // good
   function good() {}
+
+  // also good
+  const alsoGood = () => {}
   ```
 
 - [10.3](#10.3) <a name="10.3"></a> When using IIFEs, always wrap the function parentheses, with dangling parentheses for the function call.


sambostock:
The current _Why?_ section looks like this

> Why? Function declarations are named, so they’re easier to identify in call stacks. Also, the whole body of a function declaration is hoisted, whereas only the reference of a function expression is hoisted. This rule makes it possible to always use Arrow Functions in place of function expressions.

I'm not sure how to update it, especially since

>  This rule makes it possible to always use Arrow Functions in place of function expressions.

is already there... 🤔 

I could add something like 

> Arrow functions are acceptable when their `this` semantics are required, or when function expression typing semantics are desired, if working with TypeScript. For example:
> ```typescript
> interface Props { className: string }
> const Div: React.SFC<Props> = ({className, children}) => <div className={className}>{children}</div>
> ```

Thoughts @lemonmade?

PR #276, opened 2022-07-14
Fixed the URL link which led users to a HTTP 404.

Kavignon:
I was going through the docs for the code along activity related to React and observed that the link for the shared ESLint configuration at Shopify was broken. After doing some digging, I found out where it was located and wanted to help by fixing the link :)

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -1,6 +1,6 @@
 # `import {React} from 'Shopify'`
 
-This guide provides a few guidelines on writing sensible React. Many of these rules are enforced by our [shared ESLint configuration](../packages/eslint-plugin-shopify), which makes use of the excellent [eslint-plugin-react](https://github.com/yannickcr/eslint-plugin-react) project.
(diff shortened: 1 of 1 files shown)


Kavignon:
image

Result when loading the URL link from before

react/README.md
```diff
@@ -1,6 +1,6 @@
 # `import {React} from 'Shopify'`
 
-This guide provides a few guidelines on writing sensible React. Many of these rules are enforced by our [shared ESLint configuration](../packages/eslint-plugin-shopify), which makes use of the excellent [eslint-plugin-react](https://github.com/yannickcr/eslint-plugin-react) project.
+This guide provides a few guidelines on writing sensible React. Many of these rules are enforced by our [shared ESLint configuration](../packages/eslint-plugin-shopify), which makes use of the excellent [eslint-plugin-react](https://github.com/Shopify/web-configs/tree/main/packages/eslint-plugin) project.
```
mdentremont, line 3: The old URL no longer gives a 404 - but it does redirect.

Should we instead change the URL to:
```suggestion
This guide provides a few guidelines on writing sensible React. Many of these rules are enforced by our [shared ESLint configuration](../packages/eslint-plugin-shopify), which makes use of the excellent [eslint-plugin-react](https://github.com/jsx-eslint/eslint-plugin-react) project.
```

Kavignon: What's the difference between the 2 repositories?

